PR #4, opened 2015-05-26
Unified different styles in the "loop" backend

rth:
This Pull request reorganizes a bit the code in the "loop" backend of Ordinary Kriging, so the same code is used for different styles ["grid", "points", "masked"]. 
Essentially, if `style="grid"`  or  `style="masked"` we flatten the input array, compute the non masked elements in one loop as if `style="points"`, and then if necessary reshape the results back to the orginal 2D shape.   This allows a unified implementation for different input styles, removes code duplication and is in particular useful for further work on  Kriging using a moving window (so we don't re-implement the same thing several times).

This is just a minor reorganization, it should not break anything and all the test pass.  It should not affect execution speed.

Same thing could be done for the universal kriging with a loop, the 3D implementation, and to a lesser extent to vectorized version.

### DIFF
--- a/pykrige/ok.py
+++ b/pykrige/ok.py
@@ -368,7 +368,7 @@ class OrdinaryKriging:
 
         return a
 
-    def _exec_vector(self, style, a_inv, xpoints, ypoints, mask):
+    def _exec_vector(self, a_inv, xpoints, ypoints, mask):
         """Solves the kriging system as a vectorized operation. This method
         can take a lot of memory for large grids and/or large datasets."""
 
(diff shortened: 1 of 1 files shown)


pykrige/ok.py
```diff
@@ -632,8 +597,8 @@ def execute(self, style, xpoints, ypoints, mask=None, backend='vectorized'):
         if style != 'grid' and style != 'masked' and style != 'points':
             raise ValueError("style argument must be 'grid', 'points', or 'masked'")
 
-        xpoints = np.array(xpoints, copy=True).flatten()
-        ypoints = np.array(ypoints, copy=True).flatten()
```
rth: I am not sure why input arrays were flatten here before, if for instance style == 'grid'. Anyway either this could be done here or in the corresponding backend. For the moment removing it from this line, since it is handled  both by `loop` and `verctorized` backends.

rth:
I updated the "vectorized" backend of ok.py in the same way. Also the code that does all the checks regarding the arrays shape / whether it is masked or not , is factorized outside of the backend function. This allows for the backend function remain very simple, which is particularly usefull if another backend (e.g. Cython) were to be added.

I will not add more commits to this pull request, from this point on, and use this code as the basis to add moving window Ordinary Kriging.

Issue #2, opened 2015-04-19
Working with a large grid size 


rth:
Hello,

thanks you so much for working on this and making it available on github!

I would like to use this module with a moderately large grid size of the order of nx=1000, ny=1000 and with a number of points n=100.  Given that  `UniversalKriging`   solves the linear system,  `A X = b`, where `A` is, I believe, an array of shape `(nx, ny, n, n)`, this  currently requires roughly `nx*ny*n**2*8/1e9= 80 GB` of RAM simply to define that matrix in 64bit , and subsequently I'm running out of memory on my laptop. 
     
I was wondering if I was possible to optimise this code so it is more memory efficient? For instance, the option would be to  port the `execute` method in `uk.py` to Cython, so  it uses a lower number of temporary numpy arrays for indexing, etc., which would same some memory. However, at the end of the day, it's still necessary to define that same `A` matrix (which is not sparse as far as I can tell) for the linear system.  I suppose, it's not possible to do a domain decomposition on the grid (decompose it in regions and do the calculation region by region),  right?  Is there any other things that could be attempted? I can do some optimisation , I would just need some advice about where to start since I don't know much about Kriging.

Thanks,

bsmurphy:
Interesting, I hadn't actually anticipated such an issue when I redesigned the code to solve the whole system with a single massive array. But of course, such a large array would indeed take up too much memory. The previous version of the code (v 1.0.1ish) would probably have worked for you, as it actually loops through each grid point and sets up a different Ax=b system to solve each time. The problem of course is that you have to loop through each point in the grid, which in your case is quite large.

Regarding the kriging business, you're right about the A matrix, it isn't sparse. Working through subsections of the grid would be reasonable to cut down on the massive-array issue, as long as you of course incorporate all of the data each time. So you'd maybe have to loop through solving a couple of giant matrix equations with array A of shape (nx_sub, ny_sub, n, n). (Also, I think that's the right shape, although I can't recall exactly what I did right off the top of my head right now... The idea is that you have an n x n matrix defined at each of the grid points and each of those matrices is the A matrix in an Ax=b matrix equation at each grid point.) This could possibly be a good solution, as it may be faster than actually having to loop through each point in the (large) grid and still avoid using too much memory.

[...]

PR #5, opened 2015-06-03
Ordinary kriging on a moving window

rth:
This should be merged after the pull request #4  (as right now this partially redundant with it).

Ordinary kriging on a moving window (see issue #2 ) is implemented , and will be enabled if not Null `n_closest_points` is passed to the routine,

      z, ss = OK.execute(..., n_closest_points=100)
vectorized backend is not supported in this case, mostly intentionaly, since potentially the `A` martix can have variable shape, and it is therefore not suitable for vectorization.

 In terms of performance, say we take a grid of `100x100` and 100 closest points, bellow is a timing  (with the loop backend) as a function of the kriging data size,

 time to solution (sec)        |   n=100 | n=200 | n=400 | n=800 | n=1600 | n=3200 |  n=6200
 -------|-----------|------|-----|-----|-----|------|----|----
  OK                             |  0.63 |  0.798 |  1.156 |  3.76 | 15.8 | 67.49 | 289.6
   OK moving window | 3.63  | 3.56    |  3.68    | 3.72 | 4.23 | 5.32 | 12.03 
so for small data sizes, this is actually slower. However  as the data size, increases, OK on moving window execution time remains almost constant, and there is simply no other way to do kriging of say 10k  points.
As to the results, they look reasonable enough, although this migh need more testing. Another thing is `scipy.spatial.KDTree` is used to get the nearest neighbours, which is faster then calculating all the distances between the mesh and the data points.

This also adds a cython backend for both OK, and OK with kriging window, all styles (masked, grid, points) are supported, that can be called with,
    
       z, ss = OK.execute(..., backend='C')

To give an idea of the speedup, assuming, that we take 50 closest points, the speed up between the Cython version and the C version is given below (the format is `OK speed up / OK moving window speed up`),

 relative speed up        |   n=100 | n=200 | n=400 | 
 -------|-----------|------|-----
  100x100  | 3.67  / 3.68 | 2.87 /  3.80 | 1.88 / 3.54
   400x400 |  1.35/    3.25               |              1.63 /     3.57    | 1.65 / 3.64

it is not that large, but there still might be some room for improvement.


bsmurphy:
Awesome. Thanks for putting all this together, and sorry I haven't gotten back to it sooner. I'll get working on a PyPI release.

dshean:
This looks like the solution I need, as I am working with (lots of) 3D points.  Any plans to add this functionality to the 3D kriging options?

bsmurphy:
Sure, why not. It needs to be done eventually. It shouldn't be too difficult to extend @rth's pure Python implementation to 3D. Might be a few weeks, but I'll work on implementing this capability. @rth - I've also been thinking recently about adding an 'adaptive' capability, so that if the spatial covariance isn't stationary (changes across the dataset) you can automatically re-estimate the variogram model on subsets of points. I haven't thought too much about this, but I think it should fit relatively easily into your moving window implementation.

dshean:
Sounds great - thanks for considering.  The adaptive capability you mention would also be useful.

bsmurphy:
Sorry it took me a while to get to this. I've made a quick modification to the 3D ordinary kriging code to include the moving window option. I've put this update on the development branch, as I've realized there are a lot of changes I'd like to make and I'm going to keep tinkering around on this development version for a little while. The OK3D moving window change does seem to work, so give it a try. Extending to UK and UK3D is the next step.

Issue #8, opened 2015-09-08
Can't Install


luismiguelbravo:
Hello, thanks for sharing

I get de error when I try

pip install pykrige

![image](https://cloud.githubusercontent.com/assets/5140529/9741949/3200e3da-5627-11e5-9436-a5cadcf798d5.png)

rth:
Hello,

what version of anaconda are you using (32bit, 64bit) and what is your OS? Also can you copy-past the full error output instead of the screenshot?

I cannot reproduce this error so far. With a new conda install (64bit on linux), running

    conda install  numpy cython scipy
    pip install pykrige

installs the package as expected.

bsmurphy:
I'm actually also getting this error on my Windows machine (running Windows 10, 64-bit). (Only checked the updates on my Mac desktop before.) Here's the full error output from my pip debug log:

```console
Getting page https://pypi.python.org/simple/pykrige/
URLs to search for versions for pykrige in c:\python27\lib\site-packages:
* https://pypi.python.org/simple/pykrige/
Analyzing links from page https://pypi.python.org/simple/pykrige/
  Found link https://pypi.python.org/packages/source/P/PyKrige/PyKrige-0.1.1.tar.gz#md5=87c987ffa7aededb3ddbc71e352f2cf2 (from https://pypi.python.org/simple/pykrige/), version: 0.1.1
  Found link https://pypi.python.org/packages/source/P/PyKrige/PyKrige-0.1.2.tar.gz#md5=e5e3fb39a046a930c64072c919c9f100 (from https://pypi.python.org/simple/pykrige/), version: 0.1.2
  Found link https://pypi.python.org/packages/source/P/PyKrige/PyKrige-0.1.tar.gz#md5=806c966f983632741a7f8160a4492d94 (from https://pypi.python.org/simple/pykrige/), version: 0.1
  Found link https://pypi.python.org/packages/source/P/PyKrige/PyKrige-0.2.0.tar.gz#md5=4c664f9e5a949da391b59bbd1c9d19f7 (from https://pypi.python.org/simple/pykrige/), version: 0.2.0
  Found link https://pypi.python.org/packages/source/P/PyKrige/PyKrige-1.0.1.tar.gz#md5=002ffc1dbadc1cd7c0faaf726fa2df30 (from https://pypi.python.org/simple/pykrige/), version: 1.0.1
[...]

PR #9, opened 2015-09-22
Adding continuous integration

rth:
This pull request (PR) adds continuous integration on Linux (Travis CI, configuration in `.travis.yml`) and Windows (AppVeyor, configuration in `appveyor.yml`). The tests suite is run with Python 2.7, and miniconda for every new commit (both in 32 bit and 64 bit on Windows). @bsmurphy when this PR is accepted, you would need to,
  - create an account for your repo on these two free CI services
  -  update status links (see [here](https://github.com/rth/PyKrige) ) in the `README.md` with the correct links for your repo

This PR also removed the build of C extensions on Windows, and falls back to a pure Python implementation, as discussed in issue #8 . Although, this is not always necessary (for instance on AppVeyor, these extensions are built fine, since it has a C compiler installed), and this would need to be improved in the future. For instance, ideally, these extensions should not be built only if a C compiler is not present.


README.md
```diff
@@ -3,6 +3,9 @@ PyKrige
 
 Kriging Toolkit for Python
 
+[![Build Status](https://travis-ci.org/rth/PyKrige.svg?branch=master)](https://travis-ci.org/rth/PyKrige)
+[![Appveyor Status](https://ci.appveyor.com/api/projects/status/github/rth/PyKrige?branch=master&svg=true)](https://ci.appveyor.com/project/rth/pykrige)
```
rth: Url links that would need updating

setup.py
```diff
@@ -5,12 +6,33 @@
...
+               Extension("pykrige.lib.lapack",
+                         ["pykrige/lib/lapack.pyx"],
+                                   **compile_args),
+               Extension("pykrige.lib.variogram_models",
+                         ["pykrige/lib/variogram_models.pyx"],
+                                   **compile_args),]
+
+class build_ext_compiler_check(build_ext):
+    def build_extensions(self):
+        compiler = self.compiler
```
rth: Looking at attributes of `self.compiler`  might allow to determine whether a compiler is avalable...

PR #15, opened 2016-07-07
fix setup.py such that it won't crash if missing numpy/cython when using pip

MasterOdin:
Right now, if doing ```pip install pykrige``` it'll fail if you don't have Cython or numpy installed before running the pip command.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ build/
 *.swp
 .idea/*
 dist/*
+*.c
+*.so
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,6 @@ import sys
 from os.path import join
 from setuptools import setup, Extension
 from distutils.errors import CCompilerError, DistutilsExecError, DistutilsPlatformError
-
 ext_errors = (CCompilerError, DistutilsExecError, DistutilsPlatformError)
 
 NAME = 'PyKrige'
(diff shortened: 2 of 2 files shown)


rth:
@MasterOdin Thanks for contributing!

I have mixed filling about this. I agree that PyKrige should install even if Cython is not installed, that part is ok. 

However, IMO it's not such a good idea to try to install numpy/scipy automatically with pip when running `setup.py`, and it's better instead to specify them as hard dependencies for the user to install manually. See for instance the [installation instructions for sckit-learn](http://scikit-learn.org/stable/install.html). 
> We don’t recommend installing scipy or numpy using pip on linux, as this will involve a lengthy build-process with many dependencies. Without careful configuration, building numpy yourself can lead to an installation that is much slower than it should be. If you are using Linux, consider using your package manager to install scikit-learn. It is usually the easiest way, but might not provide the newest version. If you haven’t already installed numpy and scipy and can’t install them via your operation system, it is recommended to use a third party distribution.
as this could also lead to obscure BLAS/LAPACK errors (as the one currently in Travis-CI output). On Windows/Mac users should use a scientific python distribution (e.g. anaconda) anyway for performance and mixing pip with conda install is not the best idea either.

So I think that keeping a `import numpy` in `setup.py` may not be such a bad idea. At least the resulting error is straightforward and makes it user's responsibility to properly install numpy/scipy dependencies, depending on his OS & python environment.  

@bsmurphy What do you think?

MasterOdin:
I wasn't aware of that about numpy and scipy. I guess in that case, should add a hard check at the top of the setup for both of these that throws an ImportError with "Missing Numpy. Please install this for your OS before installing PyKrige. Instructions can be found at http://scipy.org/install.html" instead of just the default ImportError python would throw when it failed to load numpy.

bsmurphy:
Sorry for the delayed response... I agree with @rth on this one, I think forcing an install of the dependencies through the setup script isn't the best thing to do. Installing numpy, scipy, and matplotlib through pip can unleash all sorts of trouble (especially for matplotlib). I like the idea of throwing a custom error with directions to the user to install these dependencies separately much better.

bsmurphy:
I also do like the changes to continue installation without having Cython installed. I think that's a really good idea.

MasterOdin:
I've modified this to then have a check for the three requirements to be "hard" requirements so that it'll print out an error message and then die if missing any of them, pointing to the install page for the SciPy stack. I'm not sure why Travis is failing as conda should have all three of these packages installed already?

rth:
Not sure what's happening with travis-ci: I'm not able to login anymore and the public build page shows "The repository at bsmurphy/PyKrige was not found."  Maybe they have some service issues, let's ignore that for the moment. I manually checked that the installation works fine on Linux (and appevyor build is still ok for Windows). Looks good to me: let's merge it.

bsmurphy:
It looks like, for some reason, the check for the hard dependencies itself is failing -- in the Travis CI output, the little message that @MasterOdin has appear when __import__(*) fails is in fact appearing, even though the three base dependencies are installed in the virtual environment... 
Since @rth has confirmed that it works on Linux I'm going to go ahead and merge. We can deal with any other problems from there...
Thanks @MasterOdin! Very useful contribution.

bsmurphy:
@rth, can you see the Travis CI output from the tests on the master branch? Looks like setup.py is failing in both Python 3 and Python 2 for some weird reason that I can't easily figure out...

PR #18, opened 2016-10-13
Added Cython support for python 3

mjziebarth:
I added a workaround to replace the PyCObject_AsVoidPtr that prevented importing the cython module by PyCapsule_GetPointer. Depending on python version, setup.py compiles either lib/lapack.p* or lib/lapack_py3.p*
Code adapted from https://github.com/statsmodels/statsmodels/blob/master/statsmodels/src/capsule.h

Was so far tested only with test.py (and on python 3), which ran without errors.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,24 +90,16 @@ def run_setup(with_cython):
                                 extra_link_args=['-O2', '-march=core2', '-mtune=corei7'])
         else:
             compile_args = {}
+        
+        ext_modules = [Extension("pykrige.lib.cok", ["pykrige/lib/cok.pyx"], **compile_args),
+                       Extension("pykrige.lib.variogram_models", ["pykrige/lib/variogram_models.pyx"], **compile_args)]
+        
         # Transfered python 3 switch here. On python 3 machines, will use lapack_py3.pyx
         # instead of lapack.pyx to build .lib.lapack
         if sys.version_info[0] == 3:
-            ext_modules = [Extension("pykrige.lib.cok", ["pykrige/lib/cok.pyx"],
-                                     **compile_args),
-                           Extension("pykrige.lib.lapack", 
-                                     ["pykrige/lib/lapack_py3.pyx"], **compile_args),
-                           Extension("pykrige.lib.variogram_models",
-                                     ["pykrige/lib/variogram_models.pyx"],
-                                     **compile_args)]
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -96,9 +90,24 @@ def run_setup(with_cython):
                                 extra_link_args=['-O2', '-march=core2', '-mtune=corei7'])
         else:
             compile_args = {}
-        ext_modules = [Extension("pykrige.lib.cok", ["pykrige/lib/cok.pyx"], **compile_args),
-                       Extension("pykrige.lib.lapack", ["pykrige/lib/lapack.pyx"], **compile_args),
-                       Extension("pykrige.lib.variogram_models", ["pykrige/lib/variogram_models.pyx"], **compile_args)]
+        # Transfered python 3 switch here. On python 3 machines, will use lapack_py3.pyx
+        # instead of lapack.pyx to build .lib.lapack
+        if sys.version_info[0] == 3:
```
rth: Could you please rather use `ext_modules = [..pykrige.lib.cok.., ..pykrige.lib.variogram_models]` then append the lapack depending on this PY3 test (to avoid code duplication).

mjziebarth: The conda issue seems to persist even though enum34 has been included. Any thoughts how to fix it?

rth: Yes, sorry, I think we can just remove [lines 19-20](https://github.com/bsmurphy/PyKrige/blob/master/.travis.yml#L20) from `.travis.yml` that shouldn't be critical/necessary IMO, as we a) download the latest miniconda verison b) update it once on line 18.

mjziebarth: Seems like Conda requires enum34 to install enum34. If I understand the .yml correctly, pip is installed in [line 24](https://github.com/bsmurphy/PyKrige/blob/master/.travis.yml#L24), so we can't use it to install enum34 beforehand?

rth:
OK, Travis should be happy now. @mjziebarth Would you mind reverting latest commits as there is a conflict in `.travis.yml`. Running the following on your `master` branch,

    git reset --soft 0f129b16e64780bd2c9ccd2a3b8e9e8a5e7cbb92  # id of the "Added Cython support for python3 part2" commit
    git push -f
should fix it (I think). Or just adding another commit to revert the original state would work too. Thanks!

PR #27, opened 2016-12-16
Regression krige

basaks:


basaks:
This PR adds [regression kriging](https://en.wikipedia.org/wiki/Regression-Kriging) functionality to `PyKrige`.

rth:
This looks nice @basaks ! Just a few comments,
 - maybe call the class `RegressionKriging`  instead of `MLKriging` (so it's consistent with `OrdinaryKriging` naming etc)
  - I'm not so enthusiastic about 
     ```
     if SKLEARN_INSTALLED:
            do some imports from sklearn
     validate_sklearn()
      ```
     I mean if scikit-learn is not installed this example is not going to run anyway, so it might be best to just import the required packages directly and then if scikit-learn is not installed this will produce an `ImportError`  which is fine, as it is a more standard way of doing things than custom error messages ?
  - In the same idea, it might also be good to remove the `r2_score`, `GridSearchCV`, `train_test_split` and `SVR` imports from `compat.py` and import them directly where they are needed (for instance from within the `MLKriging.score` method). This way user can get directly the relevant ImportError error message (and google it to find a solution) instead of dealing with another level of abstraction with imports in `compat.py`. In the same idea, what `compat.py` may be intended to do is to support multiple scikit-learn version for functions that are directly used in the PyKrige library, however the examples, I think we can just run them with sklearn v0.18 (it would be less maintenance effort). What do you think?
   - `rk.py` for `RegressionKriging` /  `MLKriging` is great. Not so sure about putting there the scikit-learn wrapper class as well (but the previous location `sklearn_cv` wasn't ideal either IMO). I imaging we could leave it there for now in this dev version, but it needs a fixed location once the v1.4 is released...
   - I would remove the `logging` import and the `log = logging.getLogger(__name__)` as it's not used
   - Finally, about tests, the `Krige` scikit learn wrapper uses random data in tests, which is probably fine as those methods are extensively tested already. For `RegressionKriging` however, testing  on random data and checking that R² score > -1 is no grantee whatsoever that this actually works as intended. Maybe you could use some of the [included datasets for tests](https://github.com/bsmurphy/PyKrige/blob/master/pykrige/test.py#L32) or some more regular non random data? What do you think?
   - I have not looked in the details of the algorithm, but I don't understand why the `fit` and `predict` methods need to take as input both `X` and `lat_lon`? Say what if I have some non geographical data, I just have the coordinates (X array `[n_samples, n_coordinates]`) and the target (y array [n_samples`]) for training, it should still be possible to do `RegressionKriging` with it, right?

Would somebody be interested in doing a review of this PR, @mjziebarth @bsmurphy ?

basaks:
Hi @rth Very helpful as usual.
The easy stuff first.
I removed `logging` and replaced using `print`. On a proper release let's pursue the reverse approach :)
Used `RegressionKriging` instead of `MLKrige`. Much better, agree.
I did not like that `if SKLEARN_INSTALLED` import idea either.

Now the stuff that we need to decide:
I would love to support multiple versions of `scikit-learn`. See latest changes in `compat.py` and `rk.py`. I like the fact that the import from `compat` file allows us to support multiple versions. I just left those imports in the `compat` that have different package structure in different version of `sklearn`. See if you like it.

Although it's used in the `RegressionKriging` class, I don't like the `Krige` class in the `rk.py` either. I justleft it there as they both use `sklearn` and `RegressionKriging` uses it. Let's think keep that in mind for the next release.

Will address the `tests` comment in a future commit in the next few days.

The idea behind regression Kriging is the following: 

1. First use some form of regression (like OLS, or, a machine learning based regression) and use all other features (covariates) of your problem (except possibly x/y or lat/lon). 
2. The regression predictions will not exactly match your tagets (z).
3. The difference of your targets and machine learning prediction is what is called the `residual` in the Kriging literature. So far no Kriging has taken place. Only regression.
4. In the next step, you Krige the residuals using OK or UK, and add back to the regression prediction to improve the regression prediction. This is where lat-lon or x-y is used. 

@bsmurphy @mjziebarth Did I get the concept of regression kriging right?

basaks:
I also added a `tox.ini` file for checking different environments locally.

basaks:
@bsmurphy @mjziebarth @kvanlombeek Can you provide some confirmation of my understanding of regression Kriging and any comment on this regression kriging implementation?
I have spent literally less than 2 weeks on Kriging in my life, and that too while while doing a lot of other things.

bsmurphy:
From just a quick glance, this looks good. Just one comment for now -- looks like you generate random dummy data for the unittests. Do you have any examples for which you know the result (validated by some other code, perhaps) that you could add to the tests?

basaks:
I have added a more realistic test case for `RegressionKriging`.
I have also added a method that in the `RegressionKriging` class that clarifies how kriging is being used during regression kriging.

basaks:
@bsmurphy I have not validated it with any other software. In fact, I don't have any other software to validate against. 
The regression part does not need any validation since that is done via scikit-learn. 

The only question is that adding kriged residual to the regression output, is that the definition of regression kriging? Is it a good first order approximation of regression kriging? That is the only question here really :) It's rather simple in that regards.

basaks:
I have implemented this regression kriging in a private machine learning repository (which will be open sourced in the neat future). In addition to the functionality here, I can also do a k-fold cross validation in this code.

Here are the results of one such cross validation:

```
Plain randomforest regressor:

    "expvar": 0.6655807401089617,
    "lins_ccc": 0.7829746936588615,
    "r2_score": 0.6633098819689213,
    "smse": 0.33669011803107873

With regression kriging on the randomforest regressor :

    lins_ccc        = 0.8109455095605373
    r2_score        = 0.6899306852256306
    expvar  = 0.6940154246311385
    smse    = 0.31006931477436933

```
I can also post images of the ML prediction, the kriged residual correction, and also the regression kriged output, but I am not sure if that adds value here without a lot of context of what I am doing in that project.

basaks:
I have added an additional test using `from sklearn.datasets.fetch_california_housing` data, which has lat/lon as features. 
See how regression kriging improves r2 score by running the example in `examples/regression_kriging.py`.

rth:
@basaks Sorry for the late response. Thanks for adding better tests!
We could merge this and refactor it latter, but I still wanted to know if you are definitely sure that  `RegressionKrigging` cannot use the same api as all the other methods (and take in only `X`, `y` arrays, without the `lat_lon` input), because this would make things much more complicated and break scikit-learn API compatibilty. For instance see the [regression kriging implementation here](https://github.com/capaulson/pyKriging/blob/master/pyKriging/regressionkrige.py#L21)  that does not appear to take `lat_lon` as input (unless I missed something)?

basaks:
Hi @rth 
That seems to be a different definition/implementation of regression kriging that the [wiki regression kriging article](https://en.wikipedia.org/wiki/Regression-Kriging).

That package definition of regression kriging is just another form of kriging (like ok/uk). In their regression kriging implementation X is just 2d/3d points (see comments in the code). This is like ok/ok3d in PyKrige. I.e. their regression kriging only takes into account location.

The definition I used use for RK follows the definition of the wiki article. See how this equation of the wiki regression kriging article:

screenshot from 2017-01-08 21-52-59

The first term on the right hand side is the regression (like OLS/GLS/or any machine learning regression model) component, and the second term is the residual (kriged) component.

I spent more time during the last few days and I think my implementation/wiki article is what a lot of people understand as regression kriging.

In this implementation of RK, I use more predictor variables in addition to lat/lon(x/y). The first step is to use the predictor variables in the regression model. Then I calculate residuals. The last step is to krige the residual using OK/UK and add as the second term of the above equation.

[From this ](http://geostat-course.org/system/files/ov_Hengl_gstat.pdf)
```
If the target variable shows no correlation with the auxiliary predictors, the regression-kriging model reduces to ordinary kriging model because the deterministic part equals the (global) mean value.
```

`auxiliary predictors` in this statement are the predictor variables that I use in the regression model, then kriging takes place on the residual.

rth:
@basaks Thanks for the additional explanations! It's good to know that they are using a different implementation. 

That sound great. I have just 2 last comments, 
   - in the Readme (and tests) this PR still mentions `MLKrige` while it should mention `RegressionKriging`
   - aren't the notation `X` and `lat_lon` inversed with respect to what we are using elsewhere? For instance, in the `california_housing` example,
       * `lat_lon` contains the rows `Latitude`, `Longitude`
       * `X` is `['MedInc',   'HouseAge',   'AveRooms',   'AveBedrms',   'Population',   'AveOccup']`
     while everywhere else in this package is X is the spatial variable (i.e. lat_lon).
     Wouldn't it be more logical to rename `lat_lon` -> `X` and `X` -> `P` (as predictor, or anything else that would make sense)?

basaks:
@rth 
Well spotted and excellent suggestions.
I have addressed both.
I also removed inheritance in RegressionKriging due to [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle) violation. Updated example and tests and both you suggestions are included.

rth:
Great, looks good to me. Merging this PR, since @bsmurphy also reviewed it previously.

bsmurphy:
Thanks for merging this in; sorry I didn't get back on it sooner. I trust everything @basaks has done, it looks good to me.

PR #38, opened 2017-02-02
3D Regression Kriging

basaks:
As subject.
I can not provide the example for the 3d kriging yet, but can pending approval from my data provider.
I have a test case that checks functionality of the 3d kriging.


basaks:
Python2 test fails :(
Working on it.

pykrige/ok.py
```diff
@@ -520,7 +520,7 @@ def _exec_loop_moving_window(self, a_all, bd_all, mask, bd_idx):
 
         return zvalues, sigmasq
 
-    def execute(self, style, xpoints, ypoints, mask=None, backend='vectorized', n_closest_points=None):
+    def execute(self, style, x, y, mask=None, backend='vectorized', n_closest_points=None):
```
rth: The problem is this would break backward compatibility for someone calling this function `execute(style, xpoints=x, ypoints=y)`, so we can't make this change unfortunately.

pykrige/ok3d.py
```diff
@@ -511,7 +511,7 @@ def _exec_loop_moving_window(self, a_all, bd_all, mask, bd_idx):
 
         return kvalues, sigmasq
 
-    def execute(self, style, xpoints, ypoints, zpoints, mask=None, backend='vectorized', n_closest_points=None):
+    def execute(self, style, x, y, z, mask=None, backend='vectorized', n_closest_points=None):
```
rth: Same comment as above

pykrige/rk.py
```diff
@@ -78,30 +96,33 @@ def predict(self, x, *args, **kwargs):
...
-        if isinstance(self.model, OrdinaryKriging):
+        if isinstance(self.model, OrdinaryKriging) or \
+                isinstance(self.model, OrdinaryKriging3D):
             prediction, variance = \
-                self.model.execute('points', x[:, 0], x[:, 1],
+                self.model.execute('points',
                                    n_closest_points=self.n_closest_points,
-                                   backend='loop')
+                                   backend='loop',
+                                   **points)
```
rth: though this means that you would have to translate `x`, `y` to `xpoints`, `ypoints` here

pykrige/test.py
```diff
@@ -1451,9 +1452,12 @@ def test_krige(self):
...
                                      )
             # run the gridsearch
-            estimator.fit(X=X, y=y)
+            if m in ['ordinary', 'universal']:
+                estimator.fit(X=X[:, :2], y=y)
+            else:
+                estimator.fit(X=X, y=y)
             if hasattr(estimator, 'best_score_'):
-                assert estimator.best_score_ > -3.0
+                assert estimator.best_score_ > -20.0
```
rth: can we rather split this in 2, one for 2D `best_score > -3` and for 3D `best_score  > -20`, as otherwise this significantly lowers the threashold for 2D (so it could fail and we would not know)..

pykrige/uk.py
```diff
@@ -233,7 +233,7 @@ class UniversalKriging:
                       'spherical': variogram_models.spherical_variogram_model,
                       'exponential': variogram_models.exponential_variogram_model}
 
-    def __init__(self, x, y, z, variogram_model='linear', variogram_parameters=None,
+    def __init__(self, x, y, val, variogram_model='linear', variogram_parameters=None,
```
rth: issue with backward compatibility..

pykrige/rk.py
basaks: @rth 
Yes, it's kinda inconvenient that the original had `val` and `z` has been mixed up. And also (x, y, z) and (xpoints, y...).

Anyway I have addressed backward compatibility, but that required some patch job.

Thank you so much as always for your excellent feedback :+1:

rth:
Yes, I know we should improve those names...

PR #39, opened 2017-02-02
Changed great-circle distance function

mjziebarth:
to a more numerically stable one (based on Vincenty formula). 
Added references while doing so.

Edit: This is also the same method that geopy's great-circle distance is based off of:
[distance.py](https://github.com/geopy/geopy/blob/master/geopy/distance.py)

### DIFF
--- a/pykrige/test.py
+++ b/pykrige/test.py
@@ -1388,16 +1388,11 @@ class TestPyKrige(unittest.TestCase):
         self.assertTrue(np.allclose(ss_func, ss_lin))
 
     def test_geometric_code(self):
-        # NOTE: This test code uses precalculated reference values. If numpy
-        #       RNG changes in future versions the generated coordinate pairs
-        #       might change as well so that reference values need to be
-        #       recalculated.
         
-        # Create points uniformly distributed across the sphere:
-        np.random.seed(2022017)
-        N=200
-        lon = 360.0*np.random.random(N)
-        lat = 180.0/np.pi*np.arcsin(2.0*np.random.random(N)-1)
+        # Create selected points distributed across the sphere:
+        N=4
+        lon = np.array([7.0,   7.0,     187.0,  73.231])
+        lat = np.array([13.23, 13.2301, -13.23, -79.3])
         
         # For the points generated with this reference seed, the distance matrix
         # has been calculated using geopy (v. 1.11.0) as follows:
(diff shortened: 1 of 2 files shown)


rth:
@mjziebarth Thanks for the PR. 

Could you add another unit tests to check that we obtain the same results as in geopy? Say for some input, compute the result with geopy externally, and then write a test checking that PyKrige produces the same (hard-coded) value, then indicate in comments how that value was computed with geopy? What do you think?

mjziebarth:
@rth Yes, created some reference values and added them (including a comment how to recreate them) to the test routine.

Currently, that's 200 pseudorandom positions on a sphere created from a constant seed. Then, a distance matrix is created and some properties of that matrix are compared with the values obtained from a geopy-generated distance matrix.

Thoughts?

pykrige/test.py
```diff
@@ -1388,9 +1388,59 @@ def test_uk3d_functional_drift(self):
...
+                                   [179.39827328486047, # max
+                                    0.49781321229536524, # min != 0
+                                    3577568.5670485105, # sum
+                                    39.42686179130957, # std
+                                    90.011132985696634, # median
+                                    5.2851101415176212, # percentile
+                                    106.42435552834375,
+                                    75.241117692324522,
+                                    109.48819700842446,
+                                    40.639971012523361])
```
rth: How about computing geopy for just 10 random points (i.e. `N=10`) and then comparing that pykridge produces the same results for these points with `assert_allclose`. 10 points should be enough to check that it's working correctly, and it's conceptually easier to understand than comparing the min, max, percentile etc of d. What do you think?

mjziebarth: I first thought about that, but here's why I didn't do it: In that case, even for small N, there would be lots of data to include in test.py. Intuitively, I found that a lot of text added without much a benefit. But I'm not too sure about it, so I could add a plain reference if you wish.

rth: Aww, right is it a `NxN` matrix.. Well we can just use `N=2` then? I mean this is just to check that the output is not off by orders of magnitude, so even checking on a single value could be enough...

mjziebarth: Okay, I can do that. Would also just use predefined coordinate pairs then and remove the numpy random code.

rth:
Everything looks good to me. Thanks for the PR!

PR #40, opened 2017-02-02
fixed tox env

basaks:
added new environments and fixed py3.5 no sklearn environment


rth:
@basaks Should we "Squash and merge" PRs? Just to keep a clean git history when there are PRs with lots of additional commits due to review comments. It also makes backporting commits with `git cherry-pick` between branches easier.. What do you think?

basaks:
@rth I agree. It was a bit too late last night while I was doing those changes.
 I have just fixed that now.

PR #45, opened 2017-02-21
bug fix: 3d krige residual was wrong

whdc:
Previously, the error was being computed by comparing the kriged value to the z-axis coordinate — a corollary of misnaming the krige result `z_`.

### DIFF
--- a/pykrige/core.py
+++ b/pykrige/core.py
@@ -498,10 +498,9 @@ def find_statistics_3d(x, y, z, vals, variogram_function, variogram_model_parame
             delta[n] = 0.0
             sigma[n] = 0.0
         else:
-            z_, ss_ = krige_3d(x[:n], y[:n], z[:n], vals[:n], (x[n], y[n], z[n]),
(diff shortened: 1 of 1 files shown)


basaks:
Thanks for this @whdc 
Appreciate.

rth:
Thanks for this fix @whdc !

We should probably add more unit-tests for 3D krigging. Ideally, it would have been great to find a tests that failed before this PR and passed with it...  
However, this only affects the `delta` parameter, which does not seem to be directly used in `ok3d.py` as far as I can tell. It would also be useful to document the `find_statistics` / `find_statistics_3d` functions better, as it's hard to tell looking at the code what the `delta` output variable corresponds to..

bsmurphy:
Good catch, @whdc.
@rth -- definitely a lot of work to be done on the statistics, especially as I know there are some other things that are wrong with those calculations...
(well, I don't know exactly what is wrong, but I do know that sometimes the results are wrong)

bsmurphy:
@whdc - sure thing. Fixing the statistics is my next goal, so I might call you in for some help/input as I chew through that...

PR #47, opened 2017-02-26
Variogram Refactorization and Improvements

bsmurphy:
This PR refactors the variogram initialization routine. It also fixes some deficiencies in and optimizes the automatic variogram estimation routine.


bsmurphy:
A few other notes...

(1) In the ```__init__``` method in each of the four main classes (OK/UK 2/3D), I changed the code to something like this:
```python
self.X_ORIG = np.atleast_1d(np.squeeze(np.array(x, copy=True, dtype=np.float64)))
```
The change here is that the datatype will now be forced to be a float. Two questions... Do you think this is appropriate? I've seen problems from users before where they inadvertently fed in integers instead of real numbers and bad things happened. Kriging complex numbers is a curious question, certainly something that can expanded upon but probably not appropriate for the library as is. So, I was thinking forcing to float of some sort would be good. Also, is float64 ok, or would float32 (or something else) be better?

(2) I removed the leading underscore on some of the methods in ```core.py``` (e.g., ```_adjust_for_anisotropy``` -> ```adjust_for_anisotropy```). As I understand the underscore, it denotes a protected member of a module/class that is only used within that module/class. Since we use ```adjust_for_anisotropy``` and ```initialize_variogram_model``` in multiple other modules/classes, seemed to me that it shouldn't be 'protected'. But I'm also not as familiar as others with the standard practices with Python... thoughts?

rth:
Hi @bsmurphy,

> The change here is that the datatype will now be forced to be a float. Two questions... Do you think this is appropriate? 

Well IMO ideally it would have been better to preserve the dtype (i.e. if the input is float32, or float64 keep it that way), and only cast to float if the dtype is integer.  But that's a minor thing (and possibly less trivial thing), and we can address that later..
> Also, is float64 ok, or would float32 (or something else) be better?

+1 for using float64 by default

> I removed the leading underscore on some of the methods in core.py (e.g., _adjust_for_anisotropy -> adjust_for_anisotropy). As I understand the underscore, it denotes a protected member of a module/class that is only used within that module/class.

I think this depends on the convention used. In my understanding, a leading underscore denotes not so much a protected function/but but a semi-private one (that is designed for internal use). In practice this means that such internal metods/functions are not expected to be called by the end user, and therefore have much less constraints on backward compatibility and can be changed without warnings (as opposed to public functions, without the leading `_`). So I would be more in favour on putting an `_` on as much internal functions as possible, in order to keep a small public API as that makes refactoring and development simpler. 

Will try to review this PR in detail in the next few days.

pykrige/core.py
```diff
@@ -194,29 +191,70 @@ def _adjust_for_anisotropy(X, center, scaling, angle):
...
+        g = 0.5 * (z1 - z2)**2.
+        indices = np.indices(d.shape)
+        d = d[(indices[0, :, :] > indices[1, :, :])]
+        g = g[(indices[0, :, :] > indices[1, :, :])]
 
-    g = 0.5 * dz**2
-
-    indices = np.indices(d.shape)
-    d = d[(indices[0, :, :] > indices[1, :, :])]
-    g = g[(indices[0, :, :] > indices[1, :, :])]
```
rth: Before this was applied in all cases, and now only for `coordinates_type == 'geographic'`, is this related to the fact that `pdist` returns a flat distance vector?

pykrige/core.py
```diff
@@ -266,119 +304,105 @@ def initialize_variogram_model(x, y, z, variogram_model, variogram_model_paramet
...
-    """Function that fits a variogram model when parameters are not specified."""
+    NOTE that the estimation routine works in terms of the partial sill (psill = sill - nugget) --
+    setting bounds such that psill > 0 ensures that the sill will always be greater than the nugget...
+    """
 
     if variogram_model == 'linear':
-        x0 = [(np.amax(semivariance) - np.amin(semivariance))/(np.amax(lags) - np.amin(lags)),
-              np.amin(semivariance)]
-        bnds = ((0.0, 1000000000.0), (0.0, np.amax(semivariance)))
+        x0 = [(np.amax(semivariance) - np.amin(semivariance))/(np.amax(lags) - np.amin(lags)), np.amin(semivariance)]
```
rth: Generally speaking the previous formatting aimed at keeping the lines below 80 characters (I imagine). Not sure if we want that...

pykrige/ok.py
```diff
@@ -194,44 +208,50 @@ class OrdinaryKriging:
...
         # don't get in the way. Copies are created to avoid any problems with
         # referencing the original passed arguments.
-        self.X_ORIG = np.atleast_1d(np.squeeze(np.array(x, copy=True)))
-        self.Y_ORIG = np.atleast_1d(np.squeeze(np.array(y, copy=True)))
-        self.Z = np.atleast_1d(np.squeeze(np.array(z, copy=True)))
+        # Also, values are forced to be float... in the future, might be worth
+        # developing complex-number kriging (useful for vector field kriging)
+        self.X_ORIG = np.atleast_1d(np.squeeze(np.array(x, copy=True, dtype=np.float64)))
+        self.Y_ORIG = np.atleast_1d(np.squeeze(np.array(y, copy=True, dtype=np.float64)))
+        self.Z = np.atleast_1d(np.squeeze(np.array(z, copy=True, dtype=np.float64)))
```
rth: Yeah, let's keep a `dype='float64'` here until we come up with a better solution..

pykrige/ok.py
```diff
@@ -251,15 +272,31 @@ def __init__(self, x, y, z, variogram_model='linear', variogram_parameters=None,
...
             self.variogram_function = self.variogram_dict[self.variogram_model]
+
         if self.verbose:
             print("Initializing variogram model...")
+
+        # The 'use_psill' kwarg lets the user specify whether variogram_parameters, if specified,
+        # are given in terms of the full sill or the partial sill. The full sill is easier to work
+        # with as a user, but the code works better with the partial sill. If the 'use_psill' is False
+        # (i.e., user is feeding in the full sill), the code converts to partial sill before initializing
+        # the variogram model...
```
rth: These (and other) comments are extremely informative, and we should probably put them in a doc, as it's less likely that users would read through the docs.

pykrige/ok.py
```diff
@@ -251,15 +272,31 @@ def __init__(self, x, y, z, variogram_model='linear', variogram_parameters=None,
...
+        if variogram_parameters is None:
+            vp_temp = None
+        elif use_psill:
+            vp_temp = variogram_parameters
+        else:
+            if self.variogram_model in ['gaussian', 'spherical', 'exponential', 'hole-effect']:
+                vp_temp = [variogram_parameters[0] - variogram_parameters[2],
+                           variogram_parameters[1], variogram_parameters[2]]
```
rth: I don't understand why we need to do this now?

pykrige/uk.py
```diff
@@ -281,13 +297,26 @@ def __init__(self, x, y, z, variogram_model='linear', variogram_parameters=None,
...
+
+        # see comment in ok.py about 'use_psill' kwarg...
+        if variogram_parameters is None:
+            vp_temp = None
+        elif use_psill:
+            vp_temp = variogram_parameters
+        else:
+            if self.variogram_model in ['gaussian', 'spherical', 'exponential', 'hole-effect']:
+                vp_temp = [variogram_parameters[0] - variogram_parameters[2],
+                           variogram_parameters[1], variogram_parameters[2]]
```
rth: Still don't understand this. If it is indeed needed, maybe we could put this chunk of the code in a separate private function in `core.py`?

pykrige/core.py
basaks: yes, that's desirable from a pep8 perspective.

bsmurphy:
Thanks for the review, @rth and @basaks... You're right about the 80 char limit thing, I didn't realize that the inspection in my IDE was set to use 120 chars on a line, so I was missing that...

pykrige/core.py
```diff
@@ -194,29 +191,70 @@ def _adjust_for_anisotropy(X, center, scaling, angle):
...
-    y1, y2 = np.meshgrid(y, y, sparse=True)
-    z1, z2 = np.meshgrid(z, z, sparse=True)
-    dz = z1 - z2
+    Parameters
+    ----------
+    X: ndarray
+        float array [n_samples, n_dim], the input array of coordinates
+    y: ndarray
+        float array [n_samples], the input array of values to be kriged
+    variogram_model: str/unicode
```
bsmurphy: Yeah, probably fair... my IDE was just complaining about it so I made the change to make the yellow highlighting go away

pykrige/core.py
```diff
@@ -194,29 +191,70 @@ def _adjust_for_anisotropy(X, center, scaling, angle):
...
+    # which gives pairwise distances in a condensed distance vector (distance matrix flattened to a vector)
+    # to calculate semivariances... trick scipy.spatial.distance's pdist function into thinking that the measurements
+    # are in a 2D space by stacking with an array of zeros
     if coordinates_type == 'euclidean':
-        dx = x1 - x2
-        dy = y1 - y2
-        d = np.sqrt(dx**2 + dy**2)
+        d = pdist(X, metric='euclidean')
+        y_temp = np.vstack((y, np.zeros(y.size))).T
+        g = 0.5 * pdist(y_temp, metric='sqeuclidean')
```
bsmurphy: Good suggestion, I'll go with ```y[:, None]```
The full square matrix here previously was just kind of a stepping stone to get to the vector of unique distances, so using ```pdist``` helps to clean things up I think.

pykrige/core.py
bsmurphy: I kept the old style here for the time being because ```pdist``` doesn't natively offer a 'geographic' distance calculation as far as I know. It does accept a user-specified metric function, so eventually with some more cleaning up we can feed the spherical coordinate calculation function to pdist and skip the extra steps of making the full square distance matrix. (Again, just need a vector of unique distances between points for variogram estimation.)

pykrige/core.py
bsmurphy: Yep, didn't realize my IDE was set to 120 instead of 80... I'll fix that....

pykrige/ok.py
bsmurphy: Cool. At some point in the future it would probably be worth putting in something to only convert to float64 if the input is not floating point. I can imagine forcing everything to float64 may create some problems for some people, since its got twice the memory requirement of float32...

pykrige/ok.py
bsmurphy: Cool, I'll move it up.

pykrige/uk.py
bsmurphy: Good idea, having a separate private function will reduce redundancy (what we're going for, after all). My goal here was provide a means for the user to provide either the full sill or the partial sill. I've changed things to work with the partial sill, as that fixes some of the annoying problems with the automatic variogram estimation routine. So if the user specifies the full sill, needed some way to convert to a parameter list that contains the partial sill...

pykrige/core.py
rth: @bsmurphy Aww, that's great!

pykrige/ok.py
```diff
@@ -273,7 +310,8 @@ def __init__(self, x, y, z, variogram_model='linear', variogram_parameters=None,
                 print("Using Custom Variogram Model")
             else:
                 print("Using '%s' Variogram Model" % self.variogram_model)
-                print("Sill:", self.variogram_model_parameters[0])
+                print("Partial Sill:", self.variogram_model_parameters[0])
+                print("Full Sill:", self.variogram_model_parameters[0] + self.variogram_model_parameters[2])
```
rth: How can we have both a full sill and the partial sill? I though it could only be one or the other?

rth:
I have been thinking about the `use_psil` flag and it actually a bit complicated (works only in combination with provided `variogram_parameters` and only with a subset of the possible values). Couldn't we just get rid of `use_psil` input parameter, and provide it inside the `variogram_parameters` so everything is in one place?
For instance something along the lines of,
```
The following parameters can be provided
     linear - [slope, nugget]
     power - [scale, exponent, nugget]
     gaussian, sherical, exponential, hole-effect
          -  [sill, range, nugget] -> use_psil=True (backward compatible format)
          -  [sill, range, nugget, use_psil] 
and if non provided, parameters using the partial psil will be computed.
``` 
What do you think?

pykrige/ok.py
bsmurphy: Partial sill is just the full sill minus the nugget -- the full sill is the make value the variogram reaches, and the partial sill is how much further above the nugget (y-intercept) that variogram goes

bsmurphy:
I've been thinking about the psill thing more, and maybe it would make the most sense to use a dict instead of a list... that way the user could specify specifically 'psill' or 'sill', something like this:
```python
variogram_parameters = {'sill': 10., 'range': 5., 'nugget': 1.}
```
or 
```python
variogram_parameters = {'psill': 10., 'range': 5., 'nugget': 1.}
```

For backwards compatibility, if the user gives a list instead, then we'll automatically interpret as the full sill. @rth's idea about including a flag in the list would work as well, I guess I'm just thinking that the dict approach is more explicit...

rth:
@bsmurphy Definitely, since each parameter have a meaning, passing a dict would be even better.

bsmurphy:
@rth, how do these adjustments to the ```core ``` module look? Got rid of the ```use_psill``` kwarg and went with a joint dict/list format for the ```variogram_parameter``` kwarg, which is parsed by ```core._make_variogram_parameter_list```.

rth:
Everting looks good to me. I think it's much better with a dict.

bsmurphy:
Cool, I've updated the changelog now as well. If all looks good, I'll go ahead and merge/close. Next step is the statistics...

rth:
@bsmurphy Sounds great. (And thanks for fixing the typos :) )

PR #49, opened 2017-03-03
Use warnings module for warning about ignored anisotropy in geometric kriging.

mjziebarth:
Currently, the print warning does not benefit from the warnings functionality.
E.g. if geometric kriging is done in loops, the warning will flood the standard output.

What I did to fix this:
- added a check to see if there's really anything ignored.
- use warnings module to allow user to use standard warning handling.


rth:
Looks good to me, thank you!

P.S: we are "Squashing and merging" PRs, so in the future it's better to make PR from another branch than the master of your fork (that you could later delete)... Otherwise once this PR is merged, you would have to reset your master to the latest commit before this PR, and synchronise with the `bsmurphy/PyKrige` master. Let me know if you encounter issues with it..

mjziebarth:
Thanks for merging and the suggestion, still getting used to git!

PR #50, opened 2017-03-03
Geometric example

mjziebarth:
Added an example to show usage of OrdinaryKriging with spherical coordinates.

This also adresses the comment [comment asking for an example of error when using x/y coordinates](
https://github.com/bsmurphy/PyKrige/pull/23#issuecomment-267870839) by @basaks.

Thoughts / improvement proposals?

### DIFF
--- a/examples/krige_geometric.py
+++ b/examples/krige_geometric.py
@@ -1,4 +1,11 @@
-#!/bin/python
+"""
+Geometric example
+------------------
+
+A small example script showing the usage of the 'geographic' coordinates type
+for ordinary kriging on a sphere.
+"""
+
 from pykrige.ok import OrdinaryKriging
 import numpy as np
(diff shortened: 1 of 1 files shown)


bsmurphy:
This looks good. @rth, not too familiar with the working of Sphinx -- the doc codes you just added will automatically add this to the doc pages, right?
Just want to make sure there's nothing else that needs to be done in that respect, then we can go ahead and merge...

rth:
@mjziebarth Sorry for slow review. Looks good to me too,
could you just please add a 
```
"""
Geometric example
------------------
<Some explation here> 
"""
```
at the very top of the file (see for instance `examples/regression_kriging2d.py
`) as otherwise sphinx gallery will produce errors when running this. Thanks.

Issue #31, opened 2016-12-20
[Refactoring] N-dimenstional Kriging


rth:
Just a few thoughts about possible code refactoring in PyKrige.

Currently PyKrige has separate code that implements 2D and 3D kriging (e.g. in `ok.py` and `ok3d.py`), this result in code duplication and makes code maintenance and adding new features more difficult (they need to be added to every single file). Besides the 1D Kriging is not implemented, and it might have been nice to have some 1D examples as they are easier to visualize.

In addition, PR #24 adds a scikit-learn API to PyKrige on top of the existing `UniversalKrigging` and `OrdinaryKrigging` methods. 

A possible solution to remove the current code duplication, would be to refactor the `UniversalKrigging` and `OrdinaryKrigging` to work in N-dimensions. The simplest way of doing it would be to use something along the lines of the [scikit-learn API](http://scikit-learn.org/stable/developers/contributing.html#rolling-your-own-estimator) which will also remove the need for an additional wrapper for that. The general API could be something like,
```py
class OrdinaryKrigging(pykrige.compat.BaseEstimator):
    def __init__(self, <kriging options>):
         [..]
    def fit(X, y):
        """ Where X is an array [n_samples, n_dimensions] and y is an array [n_samples]"""
        [...]
    def predict(X):
        """ Where X is an array [n_samples, n_dimensions]
        The equivalent of the current execute(style="points", ...)
        """
        [...]
```
The case of `execute(style="masked", ...)` could be done by supporting masked arrays for `predict(X)`, while the case `execute(style="grid", ...)` can be done with a helper function,
```py
def points2grid(**points):
       """ points: a list of arrays e.g [zpts, ypts, xpts] """
       grids = np.meshgrid(*points, indexing='ij')
       return np.concatenate([grid.flatten()[:, None] for grid in grids])
````
which is mostly what is [done internally by `execute`](https://github.com/bsmurphy/PyKrige/blob/master/pykrige/ok.py#L572) at present.

This would break backward compatibility though, so a major version change would be needed (PyKrige v2). 

**Update:** the refactoring could follow the steps below,

 - [x] in `core.py` merge `adjust_for_anisotropy` and `adjust_for_anisotropy_3d` into a single private function `_adjust_for_anisotropy(X, center, scaling, angle)` where X is a `[n_samples, n_dim]` array, and all the other are list of floats and, it returns the `X_modified` array. (PR #33 )
[...]

PR #51, opened 2017-03-08
Refactor statistics calculations

bsmurphy:
This PR does an ND refactor of the statistics calculation routine. Here is a list of major changes:
* The ```krige``` and ```krige3d``` methods in the ```core``` module have been merged in ```_krige```.
* The ```find_statistics``` and ```find_statistics_3d``` methods in ```core``` have been merged into ```_find_statistics```.
* ```_find_statistics``` has been updated to deal with the potential for subsequent kriging points being too close together... Previously, the mean error could at times end up being basically zero (< 1e-10), which lead to problems in the statistics calculation. The should robustify the statistics calculation routine.

This address steps 3 and 4 in #31.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -25,8 +25,8 @@ install:
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
   - conda info -a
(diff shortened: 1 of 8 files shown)


bsmurphy:
Cool, thanks. You're much better with the nuances and details of Python than I am, so I appreciate the reviews!

PR #55, opened 2017-03-10
1D kriging example

rth:
This PR adds an example of Ordinary Kriging in 1D that aims to produce somewhat similar results to this [post on Gaussian process models](https://blog.dominodatalab.com/fitting-gaussian-process-models-python/). This example produces the following plot,

download


bsmurphy:
NICE! This is great. I'll let you merge whenever you're ready.

basaks:
Nicely done :)

PR #78, opened 2017-11-21
Allowing higher precision to be written to the ASC files.

cchristelis:
# Problem

When trying to write out geographic files with a DX of < 0.01 degrees, I got a corrupt file.

# Solution

We are now allowing users to specify the precision they want to have captured in their files.

### DIFF
--- a/pykrige/kriging_tools.py
+++ b/pykrige/kriging_tools.py
@@ -10,8 +10,8 @@ Dependencies:
     numpy
 
 Callable Methods:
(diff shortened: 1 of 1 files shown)


rth:
Thanks for the PR! 

Could you please also add some simple unit test (e.g. writing an asc file with `precision=3`) to https://github.com/bsmurphy/PyKrige/blob/master/pykrige/test.py ?

MuellerSeb:
Closing due to inactivity. Feel free to re-open. I think we will refactor the IO capabilities in the future either way.

Issue #65, opened 2017-06-29
Improve docstrings formatting


rth:
Now that [online documentation](http://pykrige.readthedocs.io/en/latest/index.html) was setup, some effort to make sure all docstrings use a correct [numpy style formatting](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) is necessary.

For instance, most methods under ["API reference"](http://pykrige.readthedocs.io/en/latest/api.html) of the documentation, are currently not very well rendered.

Any help on this would be greatly appreciated. Instructions on how to build the documentation locally can be found in https://github.com/bsmurphy/PyKrige/pull/48

bsmurphy:
I'll be honest, as I was originally making the docs I didn't really have any idea about what kind of formatting was appropriate... so this definitely needs some work. Also might be a thing where gradual changes as they come up would be best...

rth:
@bsmurphy +1 for gradual changes.  Also, it's not actually that much changes, just small formatting fixes: the docstrings are already really well documented, thanks, to your efforts )

bsmurphy:
Closed with PR #84.

PR #84, opened 2018-02-04
Updated docstring format

bsmurphy:
This PR updates the docstring formats to follow the numpy standard (as suggested in #65). It also updates the docstrings somewhat (e.g., with the copyright statements) and modifies the code in places to follow the 80 character per line limit.

### DIFF
--- a/pykrige/core.py
+++ b/pykrige/core.py
@@ -3,54 +3,60 @@ from __future__ import division
 from __future__ import print_function
 from __future__ import unicode_literals
 
-__doc__ = """Code by Benjamin S. Murphy
+__doc__ = """
+PyKrige
+=======
+
+Code by Benjamin S. Murphy and the PyKrige Developers
 bscott.murphy@gmail.com
 
-Dependencies:
-    numpy
-    scipy
+Summary
+-------
+Methods used by multiple classes.
 
-Functions:
-    _adjust_for_anisotropy(X, y, center, scaling, angle):
-        Returns X_adj array of adjusted data coordinates. Angles are CCW about
-        specified axes. Scaling is applied in rotated coordinate system.
-    _make_variogram_parameter_list(variogram_model, variogram_model_parameters):
-        Makes a list of variogram model parameters in the expected order if the
-        user has provided the model parameters. If not, returns None, which
-        will ensure that the automatic variogram estimation routine is
-        triggered.
-    _initialize_variogram_model(X, y, variogram_model,
-                                variogram_model_parameters, variogram_function,
-                                nlags, weight, coordinates_type):
-        Returns lags, semivariance, and variogram model parameters.
-        Variogram model parameters are estimated if user did not provide them.
-    _variogram_residuals(params, x, y, variogram_function, weight):
-        Called by _calculate_variogram_model.
-    _calculate_variogram_model(lags, semivariance, variogram_model,
-                               variogram_function, weight):
-        Returns variogram model parameters that minimize the RMSE between the
-        specified variogram function and the actual calculated variogram points.
-    _krige(X, y, coords, variogram_function, variogram_model_parameters,
-           coordinates_type):
-        Function that solves the ordinary kriging system for a single specified
-        point. Returns Z value and sigma squared for the specified coordinates.
-        Used in statistics calculation.
-    _find_statistics(X, y, variogram_funtion, variogram_model_parameters,
-                     coordinates_type):
-        Returns the delta, sigma, and epsilon values for the variogram fit.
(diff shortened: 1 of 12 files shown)


bsmurphy:
@rth, want to check to see if this looks more or less correct? Only big thing I'm not sure about is how to put in ReST formatting stuff to get this to appear correctly on that docs page:
```
variogram_parameters : list or dict, optional
        Parameters that define the specified variogram model. If not provided,
        parameters will be automatically calculated using a "soft" L1 norm
        minimization scheme. For variogram model parameters provided in a dict,
        the required dict keys vary according to the specified variogram
        model: ::
            linear - {'slope': slope, 'nugget': nugget}
            power - {'scale': scale, 'exponent': exponent, 'nugget': nugget}
            gaussian - {'sill': s, 'range': r, 'nugget': n}
                        OR
                       {'psill': p, 'range': r, 'nugget':n}
            spherical - {'sill': s, 'range': r, 'nugget': n}
                         OR
                        {'psill': p, 'range': r, 'nugget':n}
            exponential - {'sill': s, 'range': r, 'nugget': n}
                           OR
                          {'psill': p, 'range': r, 'nugget':n}
            hole-effect - {'sill': s, 'range': r, 'nugget': n}
                           OR
                          {'psill': p, 'range': r, 'nugget':n}
```

I used a double colon (```::```) to try to force a literal block, but not sure if that's best...

pykrige/core.py
```diff
@@ -3,62 +3,68 @@
...
+Summary
+-------
+Methods used by multiple classes.
+
+Routines
+-------
+_adjust_for_anisotropy(X, y, center, scaling, angle):
+    Returns X_adj array of adjusted data coordinates. Angles are CCW about
+    specified axes. Scaling is applied in rotated coordinate system.
+_make_variogram_parameter_list(variogram_model, variogram_model_parameters):
```
rth: Do we still want to keep these summaries of routines? Essentially that's the same thing as the [docs API page](file:///home/rth/src/PyKrige/doc/_build/html/api.html) except that it takes more effort to manually keep them up to date. Also, there are private functions, so it's not something that is supposed to appear in the Public API (i.e. summary of available functions) as users should not have to use this directly..

pykrige/ok.py
```diff
@@ -606,74 +553,95 @@ def _exec_loop_moving_window(self, a_all, bd_all, mask, bd_idx):
...
+            style : str
+                Specifies how to treat input kriging points. Specifying 'grid'
+                treats xpoints and ypoints as two arrays of x and y coordinates
+                that define a rectangular grid. Specifying 'points' treats
+                xpoints and ypoints as two arrays that provide coordinate pairs
+                at which to solve the kriging system. Specifying 'masked'
+                treats xpoints and ypoints as two arrays of x and y coordinates
+                that define a rectangular grid and uses mask to only evaluate
+                specific points in the grid.
+            xpoints : array_like, dim N
```
rth: In all the places in the docstrings where we use `dim` it might be better to replace it by `shape` (the dim is always 1, 2  or 3 here).

pykrige/variogram_models.py
```diff
@@ -3,50 +3,66 @@
...
+bscott.murphy@gmail.com
 
-Copyright (c) 2015-2017 Benjamin S. Murphy
+Summary
+-------
+Function definitions for variogram models.
+
+Routines
+--------
+linear_variogram_model(m, d)
```
rth: I forgot, one never need to call these variogram functions directly, right? 
In this case might be better to move this description directly to the docs?
For instance under `doc/variogram_functions.rst` and add `variogram_functions` to `toctree` in `index.rst`, that way it would be easier to find it.. This could also be a place to discuss more general aspects of variogram functions...

pykrige/core.py
bsmurphy: I'll move the summaries down to each individual function, they're just intended to keep track of what's what internally anyways

pykrige/ok.py
bsmurphy: Will do

pykrige/variogram_models.py
bsmurphy: Yeah, these never get directly called by the user. Moving the descriptions directly to the docs is a good idea.

rth:
It's strange that the [online docs](http://pykrige.readthedocs.io/en/latest/generated/pykrige.ok.OrdinaryKriging.html#pykrige.ok.OrdinaryKriging) didn't get updated after this PR got merged. Need to check maybe the build on ReadTheDocs failed...

bsmurphy:
I was wondering about that... thanks for checking

PR #89, opened 2018-03-16
make 'rk' importable and installable during pip install

basaks:
This PR makes `RegressionKriging` a package level importable class like the other (OK/UK)Kriging classes.

### DIFF
--- a/pykrige/__init__.py
+++ b/pykrige/__init__.py
@@ -48,5 +48,6 @@ from .ok import OrdinaryKriging
 from .uk import UniversalKriging
 from .ok3d import OrdinaryKriging3D
 from .uk3d import UniversalKriging3D
+from .rk import RegressionKriging
 
-__all__ = ['ok', 'uk', 'ok3d', 'uk3d', 'kriging_tools']
+__all__ = ['ok', 'uk', 'ok3d', 'uk3d', 'kriging_tools', 'rk']


rth:
Thanks for the  PR @basaks .

The issue is that `rk.py` imports sklearn and so importing it in the `__init__.py` would make scikit-learn a hard dependency which [it isn't now](https://github.com/bsmurphy/PyKrige/blob/bf0df599deccf257e859e02a64280538d4ed9398/setup.py#L28). I'm not sure we want to do that.

Maybe with the refactoring proposed in https://github.com/bsmurphy/PyKrige/issues/31#issuecomment-362962642 for pipelines using scikit-learn would make sense anyway. So let's at least wait before v1.4 is released before merging this. What do you think.

rth:
> Anyway, it's affecting people who are trying to use RegressionKriging. We need to come up with a way forward.

True, but,
```py
from pykrige.rk import RegressionKriging
```
is not fundamentally different from
```py
from pykrige import RegressionKriging
```

Current [examples for other classes](http://pykrige.readthedocs.io/en/latest/overview.html#universal-kriging-example) also don't use the top level module FWIW.

Let's release v1.4 and try to come up with a solution for it.

MuellerSeb:
you could add a try/except block in the ``__init__.py`` so it is only present if sklearn is importable.

MuellerSeb:
I will resolve this in a separate PR.

PR #94, opened 2018-04-23
Added a couple of methods to obtain the variogram model points or to save them to a file 

Daniel-M:
Hi, I've added a couple of methods to UniversalKriging and OrdinaryKriging:

* `get_variogram_points`: wich returns a numpy.array `[lags, variogram_value_at_lags]` which can be used of interest for the user (for further computations).   
* `save_variogram_to_file`: wich saves the variogram to a file `output_file` with separator `separator`(which defaults to space ` `) with the format `lag variogram_value_at_lag` according to the separator. Making a call with separator="," would save the values as `lag,variogram_value_at_lag`.

### DIFF
--- a/pykrige/ok.py
+++ b/pykrige/ok.py
@@ -404,19 +404,19 @@ class OrdinaryKriging:
                                         self.lags), 'k-')
         plt.show()
 
-    def get_variogram_model_points(self):
+    def get_variogram_points(self):
         """Obtain the data pairs for the variogram model.
-           It is possible to save the model to a file if output_file
-           indicates the full path to 
+           returns a numpy.array with the form [lag, variogram_value_at_lag]
         """
         variogram_points = self.variogram_function(self.variogram_model_parameters, self.lags)
 
         return np.array([self.lags, variogram_points])
 
     def save_variogram_to_file(self, output_file, separator=" "):
-        """Obtain the data pairs for the variogram model.
-           It is possible to save the model to a file if output_file
-           indicates the full path to 
+        """Save the variogram model to a file with name output_file 
+           with the form `lag variogram_value_at_lag` the separator
+           variable specifies which character will separate lags from
+           variogram values, and defaults to a single space.
         """
         variogram_points = self.variogram_function(self.variogram_model_parameters, self.lags)
 
--- a/pykrige/ok.py
+++ b/pykrige/ok.py
@@ -404,6 +404,34 @@ class OrdinaryKriging:
                                         self.lags), 'k-')
         plt.show()
 
+    def get_variogram_model_points(self):
+        """Obtain the data pairs for the variogram model.
+           It is possible to save the model to a file if output_file
+           indicates the full path to 
+        """
+        variogram_points = self.variogram_function(self.variogram_model_parameters, self.lags)
+
+        return np.array([self.lags, variogram_points])
+
+    def save_variogram_to_file(self, output_file, separator=" "):
+        """Obtain the data pairs for the variogram model.
+           It is possible to save the model to a file if output_file
+           indicates the full path to 
+        """
+        variogram_points = self.variogram_function(self.variogram_model_parameters, self.lags)
+
+        if output_file != None:
+            try:
+                with open(output_file,"w") as out:
+                    for l in range(len(self.lags)):
+                        print("{}{}{}".format(self.lags[l],
+                                              separator,
+                                              variogram_points[l]), file=out)
+            except IOError:
+                print("Couldn't save to file {}".format(output_file))
+            except:
+                print("An error has ocurred while saving {}".format(output_file))
+
     def switch_verbose(self):
         """Allows user to switch code talk-back on/off. Takes no arguments."""
         self.verbose = not self.verbose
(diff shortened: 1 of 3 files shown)


pykrige/ok.py
```diff
@@ -404,6 +404,34 @@ def display_variogram_model(self):
                                         self.lags), 'k-')
         plt.show()
 
+    def get_variogram_points(self):
+        """Obtain the data pairs for the variogram model.
+           returns a numpy.array with the form [lag, variogram_value_at_lag]
+        """
+        variogram_points = self.variogram_function(self.variogram_model_parameters, self.lags)
+
+        return np.array([self.lags, variogram_points])
```
rth: I don't really see the advantage of concatenating lags and variogram points in term of the API.

Maybe just return the `variogram_points` would be sufficient? `lags` can already be accessed as an attribute.

pykrige/ok.py
```diff
@@ -404,6 +404,34 @@ def display_variogram_model(self):
                                         self.lags), 'k-')
         plt.show()
 
+    def get_variogram_points(self):
+        """Obtain the data pairs for the variogram model.
+           returns a numpy.array with the form [lag, variogram_value_at_lag]
```
rth: The type of return variable(s) should be documented see e.g. https://github.com/bsmurphy/PyKrige/blob/4c54e28da59dccd3620e586cf38e88f56a4c5eca/pykrige/ok.py#L556

pykrige/ok.py
```diff
@@ -404,6 +404,34 @@ def display_variogram_model(self):
...
 
+    def get_variogram_points(self):
+        """Obtain the data pairs for the variogram model.
+           returns a numpy.array with the form [lag, variogram_value_at_lag]
+        """
+        variogram_points = self.variogram_function(self.variogram_model_parameters, self.lags)
+
+        return np.array([self.lags, variogram_points])
+
+    def save_variogram_to_file(self, output_file, separator=" "):
```
rth: I'm not so keen on this addition.

Assuming the above changes are made, this could be done fairly easily with,
```py
import pandas as pd

df = pd.DataFrame({'lag': model.lag,
                   'variogram_points': model.get_variogram_points()})
df.to_csv('<some-file>)
```
there is not need to include and maintain a custom CSV like format in pykrige for this..

pykrige/ok.py
Daniel-M: Well, I'll look into that, where can I read more about the API? I'd like to be in full compliance with it.

pykrige/ok.py
Daniel-M: I will add docstrings for that

pykrige/ok.py
Daniel-M: You are right, i'll take it out

pykrige/ok.py
bsmurphy: ```lags``` is a class attribute that I don't think is documented very well internally (I don't think it's mentioned in any of the docstrings...). You need the lags to do anything with the variogram points, so I actually don't mind returning them both.

Daniel-M:
I'd like to help on the docstrings side, maybe I don't know the API well, but I could help with styling or infering what each of the method does

Daniel-M:
As for AppVeyor, I figured the problem is with the conda environment for Python 3.5, I did looked what happened, but can't remember the specifics now

bsmurphy:
With the docstrings, we're trying to follow the numpy docstring convention: https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

Most of the main routines (```__init__```, ```execute```) use these formats now, so you can check them for reference...

rth:
> As for AppVeyor, I figured the problem is with the conda environment for Python 3.5, I did looked what happened, but can't remember the specifics now

We can ignore the failed Appveyor test, it's a known issue that is fixed on master https://github.com/bsmurphy/PyKrige/commit/0569194b0655af02a37909e337e1c157fa6362c5 
> Well, I'll look into that, where can I read more about the API? I'd like to be in full compliance with it.

We don't really have a write-up for API principles, so it mostly a discussion at review time. For my part I think it would make sense to adapt as much as possible of [scikit-learn's API design](http://scikit-learn.org/stable/developers/contributing.html#apis-of-scikit-learn-objects). Also if you are integeresed in general ideas about the scipy API principles, see the second part of [this presentation](https://www.slideshare.net/GaelVaroquaux/computational-practices-for-reproducible-science) particularly starting from slide 60.

Daniel-M:
Hey, sorry for the late commits, I've added a couple of tests for the method `get_variogram_points` and an extendend docstring to them. Do you agree with the name of the method `get_variogram_points` or is it explicit enough? on the other hand, I hope you like what I've made, It is kinda my first time using pytest, and tried to comply with the use of the already existing test data you have. Looking forward to know what you think.

Daniel.

bsmurphy:
Looks good to me! I think the name is good, fits with the general convention we have going. @rth might have some comments, but I'd say we can go ahead and merge once he takes a look.

bsmurphy:
@rth, can we merge this?

rth:
Sorry for the late response, yes this LGTM, merging. Thanks @Daniel-M !

PR #99, opened 2018-06-15
Ok get kriging matrix coordinates fix

michaelleerilee:
Benjamin et al.:
The construction of 'a' does not correctly account for 'geographic' coordinates in 1.4.
On the branch I'm forwarding, I've made some modifications that seem to work for me.
Regards,
Mike
P.S. I have found PyKrige quite useful in jump-starting my experiments in kriging. Nicely done.

### DIFF
--- a/pykrige/ok.py
+++ b/pykrige/ok.py
@@ -208,7 +208,7 @@ class OrdinaryKriging:
             self.Y_ADJUSTED = self.Y_ORIG
         else:
             raise ValueError("Only 'euclidean' and 'geographic' are valid "
-                             "values for coordinates-keyword.")
+                             "values for coordinates_type-keyword.")
         self.coordinates_type = coordinates_type
 
         # set up variogram model and parameters...
@@ -439,19 +439,19 @@ class OrdinaryKriging:
         print("Q2 =", self.Q2)
         print("cR =", self.cR)
 
-    def _get_kriging_matrix(self, n, xy=None, coordinates='euclidean'):
+    def _get_kriging_matrix(self, n, xy=None, coordinates_type='euclidean'):
         """Assembles the kriging matrix."""
 
         
-        if coordinates == 'euclidean':
+        if coordinates_type == 'euclidean':
             xy = np.concatenate((self.X_ADJUSTED[:, np.newaxis],
                                  self.Y_ADJUSTED[:, np.newaxis]), axis=1)
             d = cdist(xy, xy, 'euclidean')
-        elif coordinates == 'geographic':
+        elif coordinates_type == 'geographic':
             d = core.great_circle_distance_c(xy[:,np.newaxis],xy)
         else:
             raise ValueError("Only 'euclidean' and 'geographic' are valid "
-                             "values for coordinates-keyword.")            
+                             "values for coordinates_type-keyword.")            
             
         a = np.zeros((n+1, n+1))
         a[:n, :n] = - self.variogram_function(self.variogram_model_parameters,
@@ -562,7 +562,7 @@ class OrdinaryKriging:
         return zvalues, sigmasq
 
     def execute(self, style, xpoints, ypoints, mask=None, backend='vectorized',
-                n_closest_points=None,coordinates='euclidean'):
+                n_closest_points=None,coordinates_type='euclidean'):
         """Calculates a kriged grid and the associated variance.
 
         This is now the method that performs the main kriging calculation.
@@ -656,7 +656,7 @@ class OrdinaryKriging:
         n = self.X_ADJUSTED.shape[0]
         nx = xpts.size
         ny = ypts.size
-        a = self._get_kriging_matrix(n,coordinates=coordinates)
+        a = self._get_kriging_matrix(n,coordinates_type=coordinates)
 
         if style in ['grid', 'masked']:
             if style == 'masked':
@@ -717,7 +717,7 @@ class OrdinaryKriging:
             # Packed-complex version. 
             xy_data_c   = self.X_ADJUSTED + 1j*self.Y_ADJUSTED
             xy_points_c = xpts + 1j*ypts
-            a = self._get_kriging_matrix(n,xy_data_c,coordinates=self.coordinates_type)
+            a = self._get_kriging_matrix(n,xy_data_c,coordinates_type=self.coordinates_type)
             
 
         if style != 'masked':
@@ -767,7 +767,7 @@ class OrdinaryKriging:
                 bd = core.great_circle_distance_c(xy_points_c[:,np.newaxis],xy_data_c)
             else:
                 raise ValueError("Only 'euclidean' and 'geographic' are valid "
-                                 "values for coordinates-keyword.")
+                                 "values for coordinates_type-keyword.")
             
             if backend == 'vectorized':
                 zvalues, sigmasq = self._exec_vector(a, bd, mask)
--- a/pykrige/ok.py
+++ b/pykrige/ok.py
@@ -439,19 +439,19 @@ class OrdinaryKriging:
         print("Q2 =", self.Q2)
         print("cR =", self.cR)
 
-    def _get_kriging_matrix(self, n, xy=None, coordinates_type='euclidean'):
+    def _get_kriging_matrix(self, n, xy=None, coordinates='euclidean'):
         """Assembles the kriging matrix."""
 
         
-        if coordinates_type == 'euclidean':
+        if coordinates == 'euclidean':
             xy = np.concatenate((self.X_ADJUSTED[:, np.newaxis],
                                  self.Y_ADJUSTED[:, np.newaxis]), axis=1)
             d = cdist(xy, xy, 'euclidean')
-        elif coordinates_type == 'geographic':
+        elif coordinates == 'geographic':
             d = core.great_circle_distance_c(xy[:,np.newaxis],xy)
         else:
             raise ValueError("Only 'euclidean' and 'geographic' are valid "
-                             "values for coordinates_type-keyword.")            
+                             "values for coordinates-keyword.")            
             
         a = np.zeros((n+1, n+1))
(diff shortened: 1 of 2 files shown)


rth:
Thanks for your PR @michaelleerilee !

Do you mean that computing the great distance calculations is complex is much faster than doing that with `great_circle_distance` that is currently included in `core.py`? 

Maybe @mjziebarth would be able to review this?

pykrige/ok.py
```diff
@@ -743,10 +758,16 @@ def execute(self, style, xpoints, ypoints, mask=None, backend='vectorized',
...
         else:
-            bd = cdist(xy_points,  xy_data, 'euclidean')
-            if self.coordinates_type == 'geographic':
+            if self.coordinates_type == 'euclidean':
+                bd = cdist(xy_points,  xy_data, 'euclidean')
+            elif self.coordinates_type == 'geographic':
                 # Convert euclidean distances to great circle distances:
-                bd = core.euclid3_to_great_circle(bd)
+                # Note: xy_points_c & xy_data_c are in a packed-complex format
+                # Note: Using packed-complex because cdist won't work
```
rth: Could you please elaborate on this? cdist does work in ND space with N=2 that should be similar to complex, or am I missing something?

michaelleerilee: I did not see where cdist handles 'geographic' lat-lon distances. It looked like you'd have to pass a lat-lon distance function in. Using Numpy array multiplication syntax seemed an efficient workaround. The bigger issue was that in some places Euclidean distances are used where geographic distances should be used for consistency.  In fact, since this pull request, I found another place where Euclidean distances are incorrectly used regardless of the geographic flag. There's nothing wrong with core.euclid3_to_great_circle, per se.

michaelleerilee: Also at this point, it may be better to reject this pull request, since I've another [branch](https://github.com/michaelleerilee/PyKrige/tree/fit_custom_variogram_too) which has gone beyond this, treating the issue in a better way, reflecting a couple of months more study of the code. FWIW, ok.py's _get_krige_matrix also needs revision consistent with the way geographic coordinates are treated. I've not worked on the problem everywhere, just the flow that goes through the OK vectorized backend, which I needed for my work.

mjziebarth: This part of the latlon-coordinate handling should be okay in the original code.
In line [706](https://github.com/bsmurphy/PyKrige/pull/99/files/9f438d1ed556f7b547c2b2b02871ccabb0822b2e#diff-51079543933ff8bb6612fa4380cefbe3R706), the lon-lat-coordinates have been converted to 3d Euclidean coordinates, hence cdist + conversion.
This was, at the time, the least invasive change I could think of. Surely direct application of great circle distance should be better in the long run. I think the use of the KDTree for n_closest_points may have influenced my decision. If there's a way to use that with spherical coordinates, core.euclid3_to_great_circle could (and should probably) be omitted completely.

mjziebarth:
Very good spot, I missed the calculation of `a` when looking over the code. My apologies for that and thanks a lot Mike for pointing it out!
It's been a while since I was working with kriging and I'm not too familiar with the maths, but I guess this can have quite severe consequences, especially if the variogram is complex.

How should we proceed with this? Ideally I'd suggest to have a fix up on PyPI asap. If I'm not mistaken, a single change in `_get_kriging_matrix` like this
```python
if self.coordinates_type == 'euclidean':
    xy = np.concatenate((self.X_ADJUSTED[:, np.newaxis],
                         self.Y_ADJUSTED[:, np.newaxis]), axis=1)
    d = cdist(xy, xy, 'euclidean')
if self.coordinates_type == 'geographic':
    d = great_circle_distance(self.X_ADJUSTED[:,np.newaxis], self.Y_ADJUSTED[:,np.newaxis],
                              self.X_ADJUSTED, self.Y_ADJUSTED)
```
should be enough to fix the urgent error.

Since Mike already mentioned the seperate branch that makes this PR obsolete (?), maybe it would be best to have a small PR with just the bug fix and then switching to that other branch. But we could also review this one.

michaelleerilee:
I agree a focussed patch and closing the PR is the way to go, though using two different methods to calculate (the same) distance in the same system raises a caution flag for me.
My development branch needs some work before the next step. 
Off topic: One important point to consider is replacing the separate matrix inversion with an integrated matrix solver, which is more stable and should lead to better results.

mjziebarth:
I don't want to take the commit-cake so would you want to work a bit more on this branch? There are still some small things I'd like to discuss then. But also if your time is precious, I could prepare a small patch branch.

mjziebarth:
@michaelleerilee @rth (also @bsmurphy )
A bump since this has stalled a bit the last month and I think it's important to fix the bug. How shall we proceed?
Regarding this PR from my side:
- I'm curious about the complex conversion of lon/lat (does it have benefits? In my opinion, it's a little less clear to read than handling lat/lon, but that's preference, I guess.).
- We should use self.coordinates_type instead of passing coordinates (IIRC I've seen it in your more advanced branch).

In case no more work on this branch is desired, I've prepared a suggestion [branch](https://github.com/mjziebarth/PyKrige/tree/geographic_fix), but as stated previously, this PR should take precedence.

pykrige/ok.py
```diff
@@ -439,12 +439,20 @@ def print_statistics(self):
         print("Q2 =", self.Q2)
         print("cR =", self.cR)
 
-    def _get_kriging_matrix(self, n):
+    def _get_kriging_matrix(self, n, xy=None, coordinates='euclidean'):
```
mjziebarth: We can use self.coordinates_type instead of passing coordinates

pykrige/ok.py
```diff
@@ -705,6 +714,12 @@ def execute(self, style, xpoints, ypoints, mask=None, backend='vectorized',
                                         np.sin(lon_p) * np.cos(lat_p),
                                         np.sin(lat_p)), axis=1)
 
+            # Packed-complex version.
```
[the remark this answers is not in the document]
michaelleerilee: Regarding complex packing: I think there was a mapping function or technique that required a signature with a single argument. This may not be an issue any more.

pykrige/ok.py
michaelleerilee: This sounds correct.

mjziebarth:
Dear @rth, @bsmurphy,
another bump. I feel like #113 and #114 may be caused by this error. From my side, we could also merge this and clean up some of the points later on. I'd volunteer for cleaning up, but I cannot merge this, so I'd need some help from one of you. I think it's especially urgent since this error may go undetected.
Regards!

rth:
I'll try to have a more detailed look this evening.

I agree that it is important to fix this, but I'm still not sure that adding a new implementation for the great circle distance is a good idea. 

If we do add scikit-learn as a optional dependency (currently it's kind of an optional one to run some examples), it does have a fast C implementation cf https://github.com/scikit-learn/scikit-learn/issues/12552#issuecomment-437270191. That one should be more thoroughly tested (and faster) that what we would be able to do here IMO, cf https://github.com/scikit-learn/scikit-learn/pull/4458#issuecomment-99217770

michaelleerilee:
Keeping the existing great circle distance calculation is quite reasonable.
Being new to the code and working on a deadline, I didn't have time to figure out a better way to calculate and propagate the pairwise distances. Also, I'm not completely happy that there are two great circle distance calculations in the existing code, I'd prefer just one, but in practice it seems they're close enough.
Michael Lee Rilee, Ph.D.Rilee Systems Technologies LLCContact: mike@rilee.net; Phone: 703-348-8127; Cell: 240-481-3254; Fax: 703-880-7202

mjziebarth:
Thanks for the quick answers!

> I agree that it is important to fix this, but I'm still not sure that adding a new implementation for the great circle distance is a good idea.

> Keeping the existing great circle distance calculation is quite reasonable.
> Being new to the code and working on a deadline, I didn't have time to figure out a better way to calculate and propagate the pairwise distances. Also, I'm not completely happy that there are two great circle distance calculations in the existing code, I'd prefer just one, but in practice it seems they're close enough.

Okay, so I would also suggest staying with one implementation. I added the Euclidean-based distance function mainly to be least intrusive possible when I was new to the code base as well. However, by now I would agree that it would be better to use just one great circle distance function. For the ```n_closest_points``` case, one could use the great circle distance separately after the KDTree search has been performed. Then, the ```euclid3_to_great_circle``` could be removed.

@michaelleerilee I had written some code in that [branch](https://github.com/bsmurphy/PyKrige/compare/master...mjziebarth:geographic_fix) I mentioned which would be my take on doing just that. If it suits, you're welcome to copy that! 

> If we do add scikit-learn as a optional dependency (currently it's kind of an optional one to run some examples), it does have a fast C implementation cf [scikit-learn/scikit-learn#12552 (comment)](https://github.com/scikit-learn/scikit-learn/issues/12552#issuecomment-437270191). That one should be more thoroughly tested (and faster) that what we would be able to do here IMO, cf [scikit-learn/scikit-learn#4458 (comment)](https://github.com/scikit-learn/scikit-learn/pull/4458#issuecomment-99217770)

Sounds good to me. I was also thinking about optionally supporting ellipsoid great circle distance, i.e. the method by [Karney (2013)](https://link.springer.com/article/10.1007/s00190-012-0578-z). I think I've seen a rather fast python wrapper somewhere (although IIRC it's still significantly slower than even the spherical great circle distance we have). Not sure how many use cases there are where that precision gain for real-world data is needed but I guess it would be nice to have. Maybe that's something for a future milestone?

MuellerSeb:
This can be closed, right?

michaelleerilee:
I'm okay with closing it.
Michael Lee Rilee, Ph.D.Rilee Systems Technologies LLCContact: mike@rilee.net; Phone: 703-348-8127; Cell: 240-481-3254; Fax: 703-880-7202 

    On Friday, January 24, 2020, 10:58:57 AM EST, Sebastian Müller <notifications@github.com> wrote:  
 
 
This can be closed, right?

PR #106, opened 2018-08-14
add variogram_parameters to krige CV parameters dictionary

olaurendin:


pykrige/rk.py
```diff
@@ -38,7 +38,8 @@ def __init__(self,
                  nlags=6,
                  weight=False,
                  n_closest_points=10,
-                 verbose=False):
+                 verbose=False,
+                 variogram_parameters=None):
```
rth: I think it's OK to put this before `verbose` to have some consistency..

pykrige/tests/test_api.py
```diff
@@ -46,3 +50,43 @@ def test_krige():
...
+    # First dictionary with a dictionary in the place of the variogram_parameters
+    param_dict3d_1 = {"method": ["ordinary3d"], "variogram_model": ["linear"],
+                 "variogram_parameters": [{'slope': 1.0, 'nugget': 1.0},
+                                          {'slope': 2.0, 'nugget': 1.0}]
+               }
+    # First dictionary with a list of lists in the place of the variogram_parameters
+    param_dict3d_2 = {"method": ["ordinary3d"], "variogram_model": ["linear"],
+                 "variogram_parameters": [[1.0,1.0],
+                                          [2.0,1.0]]
+               }
```
rth: This is pretty extensive, but I'm not convinced we really do want to compare these two cases. In the end comparing them would be testing that "variogram_parameters" works with a list and a dict (assuming `GridSearchCV` work as expected) and that it produces the same results, which is another check.

So maybe just keep `param_dict3d_1` and check that we have the right size e.g. of `estimator1.cv_results_['param_variogram_parameters'`]` (I think, or some other key there).
And also check that the `best_param_` is the one expected..

That would simplify this test a bit.

pykrige/tests/test_api.py
```diff
@@ -46,3 +50,43 @@ def test_krige():
...
+
+    estimator1 = GridSearchCV(Krige(), param_dict3d_1, verbose=True)
+    estimator2 = GridSearchCV(Krige(), param_dict3d_2, verbose=True)
+
+    # dummy data
+    np.random.seed(42)
```
rth: better to explicitly use,
```py
rng = np.random.RandomState(42)

X3 = rng.randint(..etc)
```
to avoid affecting RNG globally.

pykrige/tests/test_api.py
```diff
@@ -46,3 +50,43 @@ def test_krige():
...
+    X3 = np.random.randint(0, 400, size=(100, 3)).astype(float)
```
rth: I know this is used in the CV example, but in retrospective, I do find kriging on integers quite strange. Could we not make this,
```py
X3 = 400. * (1 + rng.rand(100, 3))
```
more simply
?

olaurendin:
I didn't know that you could explicitly give a seed for a numpy random instance, thanks!
You will figure out that I have used decimal numbers with a decent precision when it comes to testing the best score. I think that it might be a good idea for a value test since floating numbers tend to be imprecise.

MuellerSeb:
Closing due to inactivity. Feel free to re-open.

PR #107, opened 2018-08-14
Updated readme to fix import of OrdinaryKriging

harrymd:


README.rst
```diff
@@ -60,7 +60,7 @@ First we will create a 2D dataset together with the associated x, y grids,
 .. code:: python
 
     import numpy as np
-    import pykrige.kriging_tools as kt
```
rth: `kt` is still used below, so we would also need it.

PR #116, opened 2018-11-23
Geographic fix

mjziebarth:
The fix discussed in PR #99.
I added a version bump and change log. The latter would require some more work, either we could discuss that here or I'll revert those changes.

### DIFF
--- a/pykrige/ok.py
+++ b/pykrige/ok.py
@@ -665,6 +665,11 @@ class OrdinaryKriging:
             raise ValueError("style argument must be 'grid', "
                              "'points', or 'masked'")
 
+        if n_closest_points is not None and n_closest_points <= 1:
+            # If this is not checked, nondescriptive errors emerge
+            # later in the code.
+            raise ValueError("n_closest_points has to be at least two!")
+
         xpts = np.atleast_1d(np.squeeze(np.array(xpoints, copy=True)))
         ypts = np.atleast_1d(np.squeeze(np.array(ypoints, copy=True)))
         n = self.X_ADJUSTED.shape[0]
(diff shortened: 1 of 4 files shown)


rth:
Thanks @mjziebarth ! I'll try to review in more details soon.

My main question about this is whether we can write a test that would check approximate equality of krigging in `geographic` coordinates, with the results that use `euclidean` coordinates? Possibly by taking some latitude, longitude where projecting beforehand wouldn't impact results too much?

Currently `test_ok_geometric` doesn't really check anything..

mjziebarth:
Yes, such an example should easily be possible somewhere close to the equator. Probably also close to the pole in some appropriate projection, which I think would be most important. I'll write something as soon as I have some spare time!

pykrige/tests/test_core.py
```diff
@@ -2013,3 +2013,87 @@ def test_ok_geometric():
...
+    grid_x = grid_x.flatten()
+    grid_y = grid_y.flatten()
+    grid_lon = grid_lon.flatten()
+    grid_lat = grid_lat.flatten()
+
+    # Calculate and compare distance matrices ensuring that that part
+    # of the workflow works as intended (tested: 2e-9 is currently the
+    # match for this setup):
+    d_eucl = np.sqrt(  (x[:,np.newaxis] - grid_x[np.newaxis,:])**2
+                     + (y[:,np.newaxis] - grid_y[np.newaxis,:])**2)
```
rth: Use  `scipy.spatial.cdist` -- it has a higher accuracy. Here there may be some accuracy loss I think.

pykrige/ok.py
```diff
@@ -741,13 +739,36 @@ def execute(self, style, xpoints, ypoints, mask=None, backend='vectorized',
             c_pars = {key: getattr(self, key) for key in ['Z', 'eps', 'variogram_model_parameters', 'variogram_function']}
 
         if n_closest_points is not None:
+            if self.coordinates_type == 'geographic':
+                # To make use of the KDTree, we have to convert the
+                # spherical coordinates into three dimensional Euclidean
+                # coordinates, since the standard KDTree cannot handle
+                # the periodicity.
+                # Do the conversion just for the step involving the KDTree:
```
rth: Are we sure the k-NN in Euclidean space, will produce the same neighbors as when using the great_circle distance directly? I guess they are..

pykrige/tests/test_core.py
mjziebarth: Should I still do it? That would mean adding a scipy include to test script.

pykrige/ok.py
mjziebarth: If we discard numerical errors, it should, since there is that monotonic relation between the great circle and the 3d Euclidean distance (which was also used in the ```euclid3d_to_great_circle```).

Now with finite numerical precision I'm sure it's possible to construct examples at where the two differ. I would estimate that that happens at close distances or in case distances do not differ much. I guess that those are also configurations where one would have to be careful about the choice of NN in any case.

I think a useful thing could be to determine (k+1) NN, compare the distance difference between k and k+1, and warn if they are too close - that may be useful for Euclidean as well. Maybe in another PR?

pykrige/tests/test_core.py
rth: As you wish. We use scipy in the main package anyway. Unless it's less convenient to put the input array in the right format..

mjziebarth:
Thanks for the review!
Maybe the failing tests are also related to the matrix inversion improvement possibilities Mike mentioned in the original PR, though this is just guessing. For now, choosing 2e-5 circumvents the failing.
The ```git diff master...HEAD | flake8 --diff``` did not return anything so I guess the changes are compatible?

There's also the question about the version bump, I included that in a rough proposal, but it would need finishing.

PR #125, opened 2019-05-31
GSTools covariance model as variogram_model input

MuellerSeb:
With this pull request an interface to [GSTools](https://github.com/GeoStat-Framework/GSTools) is provided. The PyKrige routines ``UniversalKriging``, ``UniversalKriging3D``, ``OrdinaryKriging`` and ``OrdinaryKriging3D`` now allow an instance of the GSTools [CovModel](https://geostat-framework.readthedocs.io/projects/gstools/en/latest/tutorial_02_cov.html) as input for ``variogram_model``. Every information needed will be set internally.

An example was added to the README and to the example folder.

PR #131, opened 2019-11-14
MAINT Fix CI setup

MuellerSeb:
This PR resolves some CI problems in Appveyor and TravisCI.

Namely:
* utf-8 problems in example scripts
* sklearn future-warning from example skript
* gstools import fail for test (See: https://github.com/bsmurphy/PyKrige/pull/125)
* xvfb as a service on TravisCI (See: https://stackoverflow.com/a/55674747/6696397)


MuellerSeb:
Damn, I created the branch from the gstools branch. Maybe you have to cherry-pick the changes.

... and the whole CI framework actually needs some refactoring.

rth:
Thanks a lot @MuellerSeb ! You could try interactive rebase to remove not needed commit https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History (or yes, cherry-picking into a separate branch)..

MuellerSeb:
somehow the examples cause, that TravisCI is failing. Can we drop running the examples, when there are tests anyway? The output is even not helping at all. I would delete this, if you agree:

    - |
        # run examples
        # scikit-learn needed by the regression kriging example
        conda install -y scikit-learn
        set -x
        for f in examples/*.py; do
           python $f 2>&1 | tee -a ~/log.txt
        done
        if grep -q "Traceback (most recent call last):" ~/log.txt; then false; else true; fi

rth:
> Can we drop running the examples, when there are tests anyway?

Well we want to make sure that examples run without errors, that's a separate concern from tests.

But I agree the current solution is not very good: I would rather build docs that would also run examples with sphinx-gallery in one of CI jobs.  I think it's fine to remove that script that manually runs examples for now.

.travis.yml
```diff
@@ -1,13 +1,13 @@
...
+  - xvfb
+env: DEPS="numpy scipy cython nose matplotlib scikit-learn"
 matrix:
     include:
     - python: "3.6"
-      env: DEPS="numpy scipy cython nose matplotlib scikit-learn"
-    - python: "3.5"                                                                                         
-      env: DEPS="numpy scipy=0.18 cython nose matplotlib scikit-learn=0.18.1"
-    - python: "2.7"                                                                                         
-      env: DEPS="numpy=1.10.4 scipy=0.17 cython nose matplotlib scikit-learn=0.17.1"
```
rth: I prefer keeping explicit version. The point is to check that oldest supported version work. If versions of numpy etc are not specified, only the latest ones will be tested.

MuellerSeb:
problem with this PR is, that it is has the gstools PR as base. I will remove gstools related stuff, so we get an updated CI workflow for the current state of pykrige.

MuellerSeb:
I had to modify the compat.py file since train_score is not returned by GridSearchCV in sklearn >v0.21 by default.

MuellerSeb:
There is still a strange error with numpy on windows for python 2. From now on I am only guessing...

MuellerSeb:
Everything working again. Including example runs.

Issue #20, opened 2016-10-26
Cannot set backend to "C"


wind-chh:
I use the latest PyKrige (1.3.0) on Fedora 23, and everythong works well. But it cannot work with backend="C". 

Following is the error message:

Warning: failed to load Cython extensions.
See https://github.com/bsmurphy/PyKrige/issues/8 
Falling back to a pure python backend...

rth:
@wind-chh PyKrige was uploaded on PiPy on 2015-10-24. Could you please try with the latest development version fom github:

      pip uninstall pykrige
      pip install git+https://github.com/bsmurphy/PyKrige.git

you would need to have the dependencies installed  including numpy, scipy, matplotlib, Cython (and possibly git).

wind-chh:
I sent a pull request. It tested OK on my Fedora 23.

rth:
Fixed in #21.

PR #147, opened 2020-04-17
Update variogram_models.py

donhalmina:
The natural variogram model combines the Gaussian model and the exponential model such that model behaves like an exponential model when alpha=1 and like a Gaussian model when alpha =2. The advantage over the exponential and gaussian model is its better adjustment of smaller lag distances.

### DIFF
--- a/pykrige/variogram_models.py
+++ b/pykrige/variogram_models.py
@@ -79,3 +79,12 @@ def hole_effect_variogram_model(m, d):
         psill * (1.0 - (1.0 - d / (range_ / 3.0)) * np.exp(-d / (range_ / 3.0)))
         + nugget
     )
+
+
+def natural_variogram_model(m, d):
+    """Natural/Stable model, m is [psill, range, nugget, alpha]"""
+    psill = float(m[0])
+    range_ = float(m[1])
+    nugget = float(m[2])
+    alpha = float(m[3])
+    return psill * (1.0 - np.exp(-3.0*(d / range_)**alpha)) + nugget


MuellerSeb:
Hey there!

Thanks for that PR. In GSTools we already provide that model with the name "stable", as suggested by Wackernagel (2003) and Webster (2007). You can already use this with PyKrige. (See the GSTools example in the Documentation)

Since we want to bring the variogram models of PyKrige and GSTools in line for the next release, this model will be included.

So I think, I'll close that PR, since we would reset if either way. Also the cython counterpart is missing in this PR.

Are you OK with that decision?

donhalmina:
Dear Sabasitan

Thanks for the email. That is fine by me

I tried the GSTools with the stable model, adapting the code a bit to read
my data but the stable variagram doesn't seem to fit very well (not
calculating the range correctly). I have been viewing the documentation and
the scripts but unable to go around it. The len_scale=1.0 (I guess it's the
range). As a result the alpha value is also erroneous.

import numpy as np
import gstools as gs
import tkinter as tk
from tkinter import filedialog
import pandas as pd
from scipy.spatial.distance import pdist, squareform


root = tk.Tk()
root.withdraw()
filepath = filedialog.askopenfilenames(parent=root, title='Choose a file')
data = pd.DataFrame()
for file in filepath:
    df = pd.read_excel(file)
    data = data.append(df)
data = data.iloc[:, :3]
data = data[~np.isnan(data).any(axis=1)]
data = np.array(data, dtype=np.float)


pwdist = squareform(pdist(data[:, :2]))
max_lag = np.int(np.ceil(np.max(pwdist)))
npoint = np.int(np.floor(max_lag/40)) #20
firstlag = max_lag / 100
lags = np.linspace(firstlag, max_lag, npoint)
bin_center, gamma = gs.vario_estimate_unstructured((data[:,0], data[:,1]),
data[:,2], lags)

fit_model = gs.Stable(dim=2)
fit_model.fit_variogram(bin_center, gamma, nugget=False)
print(fit_model)
ax = fit_model.plot(x_max=max_lag)
ax.plot(bin_center, gamma)
print(fit_model)

MuellerSeb:
Could you come up with an issue for that? Please provide a minimal working example, so we can reproduce your bug.

donhalmina:
In your code, how is the range calculated? Below you will find an image with the code line being my data and the blue line, the theoretical semivariogram using the stable variogam. Based on the graph obtained from your model, it feels like some constaints have been established for some variables (range or alpha?)

GSTool_Stable Variogram

MuellerSeb:
Which version are you using? Would you mind creating an issue in GSTools with an minimal working example and data? (at least the estimated variogram + bins exported with numpy)

donhalmina:
I started creating an issue but was able to find the solution. I had to impose a constraint/bounds on both the range and sill and that solved it. Thanks 
GSTool_Stable Variogram

MuellerSeb:
Nice to see it worked. What version of gstools are you using? We made a change in the standard boundaries, so I am curious if your version already has the updated ones.

donhalmina:
I am uisng the gstools version 1.2.1

Issue #150, opened 2020-05-04
LinAlgError: singular matrix Error


NikhilDupally:
I got this error while using Ordinary kriging.
Is there any workaround for this?

MuellerSeb:
This is related to https://github.com/GeoStat-Framework/GSTools/issues/89 and https://github.com/GeoStat-Framework/GSTools/issues/79#issuecomment-615843518

We will use the pseudo inverse in the future to calculate the kriging weight by a least-square approach, if the system is singular.

At the moment, the only thing you could try is to reduce the number of conditional points, so the kriging matrix is invertible.

MuellerSeb:
You could try PyKrige from this branch: https://github.com/GeoStat-Framework/PyKrige/tree/pseudo_inv

You can directly install it with pip:
```
pip install git+https://github.com/GeoStat-Framework/PyKrige.git@pseudo_inv
```

be sure to provide cython beforehand.

In this branch, the inverse is calculated by the pseudo-inverse.

MuellerSeb:
This should be fixed with `v1.5.1` that will be released tomorrow.
Closing for now. Feel free to re-open, if you have further questions.

WuJunxu:
I also encounter this bug with pykrige==1.60 and 1.51, python==3.9, how could I solve this problem?
With approximately 700 interp points, two Group 1 succeccfuly interpolated, but Group 2 failed with 'LinAlgError: singular matrix'.

MuellerSeb:
Did you use the speudo inverse?

WuJunxu:
> Did you use the speudo inverse?

I am sorry. What is speudo inverse?

MuellerSeb:
https://geostat-framework.readthedocs.io/projects/pykrige/en/stable/generated/pykrige.ok.OrdinaryKriging.html#pykrige.ok.OrdinaryKriging

You can set `pseudo_inv=True` in all kriging routines to solve the kriging system with more robust numerical approaches.

MuellerSeb:
Closing due to inactivity. Open a discussion for further questions.

Issue #87, opened 2018-02-19
Singular Matrix Error


shihabkhan1:
I'm trying to carry out one dimensional kriging using the python module 'pykrige' but I'm getting a few errors.

I tried to pass a zero vector as argument for the second variable.

My code is as below

            hs = np.array(np.arange(0,1,bw))
            hs0 = np.zeros(len(hs))

            OK = OrdinaryKriging(P[:, 0], P[:, 1], P[:, 2], variogram_model='linear',
                     verbose=False, enable_plotting=False)
            
            z, ss = OK.execute('grid', hs, hs0)

It gives me a linear algebra error when I run the `execute` command.

    
      File "/home/home_shihab/miniconda3/lib/python3.6/site-packages/pykrige/ok.py", line 396, in _exec_vector
        a_inv = scipy.linalg.inv(a)
    
      File "/home/home_shihab/miniconda3/lib/python3.6/site-packages/scipy/linalg/basic.py", line 976, in inv
        raise LinAlgError("singular matrix")
    
    LinAlgError: singular matrix

Can anyone please help me with this?

bsmurphy:
Sorry for the slow response. Have you tried using ``` 'points' ``` instead of ``` 'grid' ```? I suspect there's something going wrong with trying to meshgrid a 1D set of points...

Yambottle:
@bsmurphy 
I met this error as well. Probably this could help to make it work, using another way to invert the cov-matrix. https://stackoverflow.com/a/54725420/12155679
@shihabkhan1 
I don't you still need this or not, please try to reduce or increase the dimension(rows) of your input. Good Luck!

MuellerSeb:
We could implement the pseudo-inverse for the case where we have a singular matrix, that can occur frequently when the number of points is to low, or the variogram-model is ill-choosen.

MuellerSeb:
I'll implement this in v1.5.1.

MuellerSeb:
You could try PyKrige from this branch: https://github.com/GeoStat-Framework/PyKrige/tree/pseudo_inv

You can directly install it with pip:
```
pip install git+https://github.com/GeoStat-Framework/PyKrige.git@pseudo_inv
```

be sure to provide cython beforehand.

In this branch, the inverse is calculated by the pseudo-inverse.

MuellerSeb:
This should be fixed with `v1.5.1` that will be released tomorrow.
Closing for now. Feel free to re-open, if you have further questions.

PR #151, opened 2020-05-05
Use pseudo inverse to prevent singular matrices

MuellerSeb:
This is adressing #87, #150 and others.
The pseudo-inverse eventually solves the system by the least square error solution of the kriging equation. This helps to solve over-/under-determined systems.

If there are redundant points, their values will be averaged in the resulting krige field.

This feature can be pluged in via a switch in the init routines of Ordinary and Universal kriging in 2D/3D.

### DIFF
--- a/pykrige/uk.py
+++ b/pykrige/uk.py
@@ -15,7 +15,7 @@ References
 ----------
 .. [1] P.K. Kitanidis, Introduction to Geostatistcs: Applications in
     Hydrogeology, (Cambridge University Press, 1997) 272 p.
-.. [2] N. Cressie, Statistics for spatial data, 
+.. [2] N. Cressie, Statistics for spatial data,
    (Wiley Series in Probability and Statistics, 1993) 137 p.
 Copyright (c) 2015-2020, PyKrige Developers
 """
@@ -29,6 +29,7 @@ from .core import (
     _initialize_variogram_model,
     _make_variogram_parameter_list,
     _find_statistics,
+    P_INV,
 )
 import warnings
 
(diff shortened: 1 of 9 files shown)


MuellerSeb:
And another note:
The example uses a linear model, that is not a valid one in 2D.

rth:
Are we confident that the matrix we are trying to invert is always Hermetian/symmetric? 

Also at least in `_exec_vector` it looks like we could solve the linear system directly with `scipy.linalg.solve` which should be faster than inverting the matrix, or am I missing something?

MuellerSeb:
Yepp, the symmetry of the kriging matrix is a property we can use. `pinvh` is a lot faster than `pinv` for big matrices.

We could also solve it, but we should use [`scipy.linalg.lstsq`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lstsq.html#scipy.linalg.lstsq), to guarantee a unique solution (should coincide with the pseudo-inverse solution)

I could check, if that solves the problem above.

MuellerSeb:
Problem is:
`linalg.solve` and `linalg.lstsq` don't support masked arrays. So this would need more refactoring.

MuellerSeb:
For some strange reason, one test is failing with `pinvh` but succeeds with `pinv`... so there are some numerical errors occurring. I would stick to `pinv` then.

@rth since the inversion in `_exec_vector` and `_exec_loop` is only done once, I think it would be ok to not use `linalg.solve`.

I would add a switch to use the pseudo-inverse, so the current behavior stays the same and user can enable the pseudo-inverse, if they need it.

MuellerSeb:
I will implement a switch `inv` in the `execute` method to select the used method.
Either:
- `inv="inv"` (default) use the inverse as before
- `inv="pinv"` use `pinv` from scipy which uses `lstsq` to compute the pseudo-inverse
- `inv="pinv2"` use `pinv2` from scipy which uses singular value decomposition to compute the pseudo-inverse
- `inv="pinvh"` use `pinvh` from scipy which uses the eigenvalues to compute the pseudo-inverse

Then everything will stay as it is and the users can select other routines for the inversion.

MuellerSeb:
I now decided to add two new keywords to the Kriging class `__init__` routines, following https://github.com/GeoStat-Framework/GSTools/pull/97:
```
    pseudo_inv : :class:`bool`, optional
        Whether the kriging system is solved with the pseudo inverted
        kriging matrix. If `True`, this leads to more numerical stability
        and redundant points are averaged. But it can take more time.
        Default: False
    pseudo_inv_type : :class:`int`, optional
        Here you can select the algorithm to compute the pseudo-inverse matrix:

            * `1`: use `pinv` from `scipy` which uses `lstsq`
            * `2`: use `pinv2` from `scipy` which uses `SVD`
            * `3`: use `pinvh` from `scipy` which uses eigen-values
```

So the current behavior is conserved and the usage of the pseudo-inverse can be opted in.

@rth : Are you OK with that?

MuellerSeb:
@rth ping. :-)

MuellerSeb:
I think, I'll just merge this by next week!

pykrige/uk.py
```diff
@@ -174,19 +175,32 @@ class UniversalKriging:
...
+        and redundant points are averaged. But it can take more time.
+        Default: False
+    pseudo_inv_type : :class:`int`, optional
+        Here you can select the algorithm to compute the pseudo-inverse matrix:
+
+            * `1`: use `pinv` from `scipy` which uses `lstsq`
+            * `2`: use `pinv2` from `scipy` which uses `SVD`
+            * `3`: use `pinvh` from `scipy` which uses eigen-values
+
+        Default: `1`
```
rth: I think it would be more explicit for the user

```suggestion
    pseudo_inv_type : str, default='pinv'
        Here you can select the scipy algorithm to compute the pseudo-inverse matrix:

            * ``pinv`` uses `lstsq`
            * ``pinv2`` uses `SVD`
            * ``pinvh`` uses eigen-values
```

rth (approved):
Thanks @MuellerSeb , sorry for late response. Generally LGTM  but I haven't done a very thorough review. 

One comment is that it might be better to use string options for pseudo_inv_type, as `UniversalKrigging(pseudo_inv_type='pinvh')` is more explicit/readable than `UniversalKrigging(pseudo_inv_type=3)`.

pykrige/uk.py
MuellerSeb: Done.

PR #157, opened 2020-05-25
Fix CI with scikit-learn related fixes

### LINKED ISSUE #154: sklearn 0.23 argument issue
https://github.com/GeoStat-Framework/PyKrige/blob/1a889a7f087ac7c3e852f6e48601f4a60efca260/pykrige/compat.py#L14

This line is not working with the newest version of sklearn, resulting in a missing `mean_train_score`.

I think, this is because they implemented keyword-only arguments there:
https://github.com/scikit-learn/scikit-learn/blob/fd237278e895b42abe8d8d09105cbb82dc2cbba7/sklearn/model_selection/_search.py#L1174

rth:
Closes https://github.com/GeoStat-Framework/PyKrige/issues/154

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
     - CIBW_BEFORE_BUILD="pip install numpy==1.17.3 scipy==1.3.2 cython==0.29.14 setuptools"
     - CIBW_TEST_REQUIRES="pytest scikit-learn gstools"
     - CIBW_TEST_COMMAND="pytest -v {project}/tests"
+    - OMP_NUM_THREADS=2
 
 before_install:
   - |
--- a/pykrige/compat.py
+++ b/pykrige/compat.py
@@ -1,7 +1,6 @@
 # coding: utf-8
 # pylint: disable= invalid-name,  unused-import
 """For compatibility"""
-import inspect
 from functools import partial
 
 
@@ -11,12 +10,6 @@ try:
     from sklearn.model_selection import train_test_split
 
     SKLEARN_INSTALLED = True
-    arg_spec = inspect.getfullargspec(GridSearchCV)[0]
-    # https://stackoverflow.com/a/56618067/6696397
-    if "return_train_score" in arg_spec:
-        GridSearchCV = partial(GridSearchCV, return_train_score=True)
-    if "iid" in arg_spec:
-        GridSearchCV = partial(GridSearchCV, iid=False)
 
 except ImportError:
     SKLEARN_INSTALLED = False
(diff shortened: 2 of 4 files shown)


pykrige/compat.py
```diff
@@ -11,12 +10,6 @@
     from sklearn.model_selection import train_test_split
 
     SKLEARN_INSTALLED = True
-    arg_spec = inspect.getfullargspec(GridSearchCV)[0]
-    # https://stackoverflow.com/a/56618067/6696397
-    if "return_train_score" in arg_spec:
-        GridSearchCV = partial(GridSearchCV, return_train_score=True)
```
rth, line 17: return_train_score was added in 0.19 while latest release is 0.23 so I think it's safe to assume the scikit-learn version used has it.

Generally I'm not keen on this re-exporting of third party modules with different default parameters, I think we should just remove this.

pykrige/compat.py
```diff
@@ -11,12 +10,6 @@
...
-    if "iid" in arg_spec:
-        GridSearchCV = partial(GridSearchCV, iid=False)
```
rth, line 19: iid is deprecated and will be removed in the next version.

.travis.yml
```diff
@@ -11,6 +11,7 @@ env:
     - CIBW_BEFORE_BUILD="pip install numpy==1.17.3 scipy==1.3.2 cython==0.29.14 setuptools"
     - CIBW_TEST_REQUIRES="pytest scikit-learn gstools"
     - CIBW_TEST_COMMAND="pytest -v {project}/tests"
+    - OMP_NUM_THREADS=2
```
rth, line 14: Avoid CPU over-subscription.

MuellerSeb: Cool.

pykrige/compat.py
MuellerSeb: That should be ok since we require `sklearn>=0.19`:
https://github.com/GeoStat-Framework/PyKrige/blob/4a974fd5216779e443914201bce424d59a3dc4b8/setup.py#L103

pykrige/compat.py
MuellerSeb: Since when it was deprecated?

rth: > https://scikit-learn.org/0.20/modules/generated/sklearn.model_selection.GridSearchCV.html

Changed in version 0.20: Parameter iid will change from True to False by default in version 0.22, and will be removed in 0.24.

> https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html

Deprecated since version 0.22: Parameter iid is deprecated in 0.22 and will be removed in 0.24

It's False by default for recent versions, so there is no need to set it to false. 0.20 was released in 2018. I don't think we need to support 0.19 (or maybe even more than a couple of scikit-learn releases really).

MuellerSeb: I am fine with that.

MuellerSeb (approved):
Looks good!

PR #158, opened 2020-07-16
Grid search update

### LINKED ISSUE #128: Is it possible to use GridSearchCV with the class OrdinaryKriging?
Hi,

I am interested in using `pykrige.ok.OrdinaryKriging` and `pykrige.uk.UniversalKriging` in many different set ups with respect to the accepted parameters. Thus, it makes sense to use `GridSearchCV` to define a big dictionary with all the parameters and let the library do the rest. However, it seems that GridSearchCV is coupled to the RegressionKriging package and not accepting parameters from OK/UK. Thus, if I use the code below specifying some of the OK parameters (e.g. variogram_parameters, anisotropy_scaling), the following error is raised:

```python
ValueError: Invalid parameter anisotropy_angle for estimator Krige(method='ordinary', n_closest_points=10, nlags=6, variogram_model='linear',
      verbose=False, weight=False). Check the list of available parameters with `estimator.get_params().keys()`.
```
Find below the code used. I wonder whether there is a way of using GridSearchCV also with the OK/UK parameters that I might have overlooked, or else the solution is to generate in a loop the models with the parameters required for this experiment.

Thanks for the great work! :-)

```python

import numpy as np
from pykrige.rk import Krige
from pykrige.compat import GridSearchCV

methods = ["ordinary"]
vario_models = ["linear", "exponential", "gaussian", "spherical", "power"]
weights = [True, False]
ani_scalings = [1]
ani_angles = [0]
verb = [True]
enable_plot = [False]
enable_stat = [False]
coordi_type = ["euclidean"]

param_dict = {  "method":methods,
                "variogram_model":vario_models,
                "weight":weights,
                "anisotropy_scaling":ani_scalings,
                "anisotropy_angle":ani_angles,
                "verbose":verb,
                "enable_plotting":enable_plot,
                "enable_statistics":enable_stat,
                "coordinates_type":coordi_type
                }

estimator = GridSearchCV(Krige(), param_dict, verbose=True)

print("List of params: ")
print(estimator.get_params().keys())

# dummy data
X3 = np.random.randint(0, 400, size=(100, 2)).astype(float)
y = 5 * np.random.rand(100)

# run the gridsearch
estimator.fit(X=X3, y=y)

print(estimator.cv_results_)
```

MuellerSeb:
This PR updates the sklearn wrapper and the Regression Kriging class in order to be compatible with all kriging routines.
resolves #128


pykrige/rk.py
```diff
@@ -43,19 +79,55 @@ def __init__(
...
+        self.drift_terms = drift_terms
+        self.point_drift = point_drift
+        self.external_drift = external_drift
+        self.external_drift_x = external_drift_x
+        self.external_drift_y = external_drift_y
+        self.functional_drift = functional_drift
         self.model = None  # not trained
         self.n_closest_points = n_closest_points
         self.method = method
+        self.val_kw = "val" if self.method in threed_krige else "z"
```
rth: This would technically fail scikit-learn's validation check in [`check_no_attributes_set_in_init`](https://github.com/scikit-learn/scikit-learn/blob/fcba0d6c0b7a6af666d6fc0978b849251b53ebcc/sklearn/utils/estimator_checks.py#L2425). It might be better to set it in fit, and also maybe make it private `_val_kw` if users are not expected to interact with it?

pykrige/rk.py
```diff
@@ -64,28 +136,37 @@ def fit(self, x, y, *args, **kwargs):
...
+            anisotropy_angle_x=self.anisotropy_angle_x,
+            anisotropy_angle_y=self.anisotropy_angle_y,
+            anisotropy_angle_z=self.anisotropy_angle_z,
+            drift_terms=self.drift_terms,
+            point_drift=self.point_drift,
+            external_drift=self.external_drift,
+            external_drift_x=self.external_drift_x,
+            external_drift_y=self.external_drift_y,
+            functional_drift=self.functional_drift,
+        )
```
rth, line 190: At this point maybe,
```
add_setup = {key: getattr(self, key) for key in ["anisotropy_scaling", ...]}
```
but writing is explicitly as done above is also fine..

rth (commented):
Thanks @MuellerSeb , generally looks good but I'm a bit concerned by the number of added parameters. Can't we make for instance,
```
        anisotropy_angle_x=0.0,
        anisotropy_angle_y=0.0,
        anisotropy_angle_z=0.0,
```
just,
````
anisotropy_angle = (0, 0, 0)
````
that would also be easier to generalize to 1D/2D/3D?

File: pykrige/rk.py (hunk 2 above)
Comment (MuellerSeb) [reply pykrige/rk.py:190]: Following the Zen of Python: Explicit is better than Implicit ;-)

File: pykrige/rk.py (hunk 1 above)
Comment (MuellerSeb) [reply pykrige/rk.py]: Will do so.

**Comment by MuellerSeb** [comment]
> Thanks @MuellerSeb , generally looks good but I'm a bit concerned by the number of added parameters. Can't we make for instance,
> 
> ```
>         anisotropy_angle_x=0.0,
>         anisotropy_angle_y=0.0,
>         anisotropy_angle_z=0.0,
> ```
> 
> just,
> 
> ```
> anisotropy_angle = (0, 0, 0)
> ```
> 
> that would also be easier to generalize to 1D/2D/3D?

This is indeed true and it is the way it is done in GSTools for example. For PyKrige v2 we should definitly do it this way and since I want to use the GSTools CovModel then, it will be simplified either ways.
For the moment I would keep it this way, since the  signatures of the kriging classes are this way. In most cases these parameters can be ignored and the class can be used in the way it was before. So for the `1.5.1` I would keep it.

**Comment by rth** [comment]
> For the moment I would keep it this way, since the signatures of the kriging classes are this way.

Right, but introducing a new parameter knowing that it will be deprecated in next version and users will have to change is not ideal. If Krigging class was another subtype of krigging I would agree about API consistency but here it's aims to be a general top level abstraction for krigging. I think it would be better to directly do small improvements in the direction of v2 for this class, rather than break things in a major way between v1 and v2 if we can help it.

**Comment by MuellerSeb** [comment]
Then I will do this for anisotropy scaling and angles. Thanks for your thoughts on that!

**Comment by rth** [comment]
Any chance we could also aggregate,
```
        self.external_drift = external_drift
        self.external_drift_x = external_drift_x
        self.external_drift_y = external_drift_y
```
into one parameter?

MuellerSeb:
> Any chance we could also aggregate,
> 
> ```
>         self.external_drift = external_drift
>         self.external_drift_x = external_drift_x
>         self.external_drift_y = external_drift_y
> ```
> 
> into one parameter?

That is true, I can do that. We could call it `ext_drift_grid`.
But there are also still some parameters left:
- `exact_values` (all routines)
- `specified_drift` (`uk` and `uk3d`) -> but this one is a bit delicate, since it would need the `specified_drift_arrays` in the `execute` method, which is hard to manage in when random samples are used for training. I'm not familiar enough with sklearn to know if this could be done. @rth : should we skip it then?

MuellerSeb:
@rth ping. :-)

MuellerSeb:
I think, I'll just merge this by next week!

rth (approved):
Thanks @MuellerSeb , LGTM.

PR #182, opened 2021-03-31
Fix sample error in 02_kriging3D.py

Rhilip:
the ax4 seems use the wrong data field, change it from `k3d3` to `k3d4`


MuellerSeb:
@Rhilip good catch! Thanks for this fix!

MuellerSeb (approved):
Looks good.

PR #188, opened 2021-04-05
typo in Geostatistics reference title

scottstanie:
noticed this when searching for the reference


MuellerSeb (approved):
Good catch, thanks! :wink:

PR #290, opened 2024-05-27
Add numpy2 support

MuellerSeb:
Since numpy 2 is around the corner (https://github.com/numpy/numpy/issues/24300), we need to be prepared.

numpy 2.0rc1 is already available and can be used as a build time dependency, so we will use this to build our wheels in the CI.

I removed support for 32bit wheels and also removed the separately built universal2 wheels for macOS, since we already provide wheels for intel and arm64.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -22,15 +22,15 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Set up Python 3.8
+      - name: Set up Python 3.9
         uses: actions/setup-python@v2
         with:
-          python-version: 3.8
+          python-version: 3.9
 
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install black 'isort[colors]<6'
+          pip install -v --editable .[lint]
 
       - name: black check
         run: |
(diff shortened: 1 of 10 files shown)


jakirkham:
Thanks Sebastian! 🙏

In order to ensure the wheels are compatible with NumPy 2, would recommend building with the NumPy 2 RC. The wheels built this way will work with NumPy 1 & 2

Just looking at the CI scripts, it appears these may be built with NumPy 1, which wouldn't be NumPy 2 compatible. Though please let me know if I've just missed something about how the builds work here

MuellerSeb:
Hey there,

IMHO the wheels should be built with np 2rc. The important lines are in the pyproject.toml:
```
     "numpy>=2.0.0rc1,<2.3; python_version >= '3.9'", 
     "scipy>=1.13.0,<2; python_version >= '3.9'", 
     "oldest-supported-numpy; python_version < '3.9'", 
     "scipy>=1.3.2,<2; python_version < '3.9'",
```

Since the wheels are built with cibuildwheel, these versions should be used.

maybe you confused this with the versions in the testing matrix?

jakirkham:
Ah ok. Maybe I did. Thanks for the clarification Sebastian! 🙏

PR #298, opened 2024-10-17
Update ok.py

SimonKraatz-USDA:
change OK.get_variogram_points() to also provide the semivariogram result. 

instead of 
lagpts , modelpts = OK.get_variogram_points()
it is now
lagpts , semivariancepts, modelpts = OK.get_variogram_points()

### DIFF
--- a/src/pykrige/ok.py
+++ b/src/pykrige/ok.py
@@ -583,6 +583,7 @@ class OrdinaryKriging:
         """
         return (
             self.lags,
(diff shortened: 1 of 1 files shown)


SimonKraatz-USDA:
In my view, the OK.get_variogram_points() should also provide the semivariance binned result (i.e., the red stars on the figure produced by OK.display_variogram_model() / or via enable_plotting=True)

Thanks!

PR #302, opened 2025-01-13
Add enable_statistics option to uk.py

cvelascof:
Add enable_statistics option to UK to avoid unnecessary computations in KED. It is done in the same way than OK.py


cvelascof:
Hi PyKrige Team,
thanks for your project. Here is a proposed change to add enable_statistics flag to uk.py to avoid unnecessary computations of statistics in uk.py when not needed.

PR #307, opened 2025-06-17
Drop Python 3.8, depend on scipy>=1.6.0, test with Python 3.13

mwtoews:
Python 3.8 reached it's end-of-life in 2024-10-07 ([PEP 569](https://peps.python.org/pep-0569/)), so should be dropped. Add Python 3.13 to the testing matrix.

Change SciPy dependency to >=[1.6.0](https://pypi.org/project/scipy/1.6.0/) to overlap with Python 3.9. The minimum version of scikit-learn is increased to overlap with Python 3.9.

This PR is similar to https://github.com/GeoStat-Framework/GSTools/pull/386

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -76,17 +76,15 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-13, macos-14]
         # https://github.com/scipy/oldest-supported-numpy/blob/main/setup.cfg
         ver:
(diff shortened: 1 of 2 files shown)


MuellerSeb:
Thanks for the effort! I really appreciate that! We need some further work on the ci to upgrade versions of github actions for up- and download of artifacts like in https://github.com/GeoStat-Framework/GSTools/pull/375.

MuellerSeb:
Superseded by #308

